Issue #21, opened 2017-12-17
How to treat  multiple valid results


julik:
One of the goals we set was to be able to recognize a file as having 2 valid formats. I wonder how this could look API wise for the customer end of the library. For example, should it be a sum type?

```
info = FormatParser.parse(io)
# What should "info" be? Should it
# be an Array of file information structs
# returned by the parsers? How do we
# make case statements on this `info`
# more convenient?

case info
when :unknown # Should we implement a null object FileInformation?
when :video # What do we do if 2 parsers manage to open the file - like .mov + .mp4?
...
```

We can also make it a Hash per type, but then again - it is likely that you will get multiple `:image` natured information structs and no `:audio` natured structs.

We can also experiment with a very smart "sum type" like umbrella FileInformation that gets returned, with something like this:

```
info = FormatParser.parse(io)
info.natures #=> [:video, :image]
info.video #=> what should this return?
```

Another question is of course performance-focused - does it make sense to allow all the parsers to have a go at the file or is it reasonable to stop at the first N matches? Currently we return right on the first match, but the parsing ambiguities is one of the things we want to explore.

tehAnswer:
**🖖**

I'm doing this one.

bermannoah:
Closed in #47

PR #37, opened 2018-01-01
Enabling FormatParser to handle different results.

tehAnswer:
_**DISCLAIMER**: This pull request is goddamn huge. I ran rubocop and moved tons of files around. If you rather review a shorter version of it, feel free to ask for it. Thank you!_

## Changelog

- `FormatParse.parse(io)` now returns an object of type `FormatParser::Result`, which acts as a wrapper object for the results of the different types of parsers.
- Added model entities `Audio`, `Document` (scaffold), `Image` and `Video` (scaffold).
- Improvements in the code styling via partially enforcing Rubocop's autocorrect.
- Major changes in the file organization for the parsers.
- Removed FormatParser mutex used to register parsers in favor of a simple array.

# Rumblings

Steemed from #21, what I believe is that the actual need is for the parser to be able to return different results based on their natures. With that understanding in mind, my proposed solution turns around what, how I see it, a Strategy-ish design pattern where the parsers (concrete strategies) are divided into _families_. The following images represent a possible logical organization in the future:

bahabah

When a file is sent to be analyzed, each family forwards it to their strategies, returning the first coincidence or nil otherwise. Although I have not done this process concurrent or parallel not to make this PR bigger in size, I would strongly recommend to give it a go in this case.

Then, once all the families have processed the file, if there is a meaningful result, a wrapper object is created which contains references to those. Otherwise, the gem returns nil as before.
 
I think this approach goes well with the enlisted goals in the README, as it keeps the code modular and thereby easy to extend, but it also lead me to add some degree of code repetition -- see `AudioParser` and `VideoParser`, they are almost identical! I could have used composition or inheritance to DRY 'em up, but honestly, I don't quite see that the gain is that much for doing it. However, there are two major pain points.

The first one of them is that this approach enforces a hidden constraint on the parsers: now a parser can only return object matching its designated nature. This shouldn't be a biggie, but for sure there'll be the situation where a file of a given could be perceived as two different natures. The easiest (yet ugliest) solution is to add a new artificial _family_.

The second one comes down to the situation when there is more than one valid result in a _family_. Historically, this code has returned with the first coincidence given its strong performance requirements, but should it? In case it shouldn't, it could lead to some breaking changes to the interface of the gem, which could be, of course, avoidable. Maybe it could be implemented as two methods, `parse` and `deep_parse`.


bermannoah:
Hey @tehAnswer, wow! Thanks for doing this. Quite a lot of work in this PR, so I was wondering: would you be okay with separating the style / rubocop changes that are not specific to the issue at hand into their own, separate PR?  I've got comments on the individual changes, but I want to make sure I'm not commenting on a rubocop suggestion. 😃

Plus -- I'd like to see how you tackled #21 specifically -- and I have a _suspicion_ 😉 that @julik will want to tinker with the style changes as well.

tehAnswer:
@bermannoah I've just discarded those changes, after all, I just used the default settings. So there's no need for another PR,  though I'd create an **_issue_** for sure!  😂  

About the rest, I can't simplify it anymore -- there's a lot of noise from moving one file to another directory. That mostly affects the actual parsers, which haven't changed much rather than their return value and namespace.

julik:
The classic inheritance approach musings, ❤️  this!

## Design

So - recently I've discovered a case where you have a single parser that can produce both image, video _and_ audio natures, and they are all valid - see #36 which should work for .MP4, .M4A, .MOV. Also there is this nice thing that you can actually have an audio-only .MOV and an audio-only .MP4.

What I am actually curious about is how we should deal with the API that the consumer of the library will see. Imagine we have a CR2 which actually _is_ both a TIFF and a... CR2. What should the metadata structure be? And how should the consumer of the library use that structure to make useful decisions based on it (an example of a useful decision is: "should I use libtiff or dcraw to load this file's pixbufs?"

Do I understand it correctly that we have something like this?

```
maybe_result = FormatParser.parse(io)
if maybe_result.video?
  maybe_result.fps #=> Float
end
if maybe_result.audio?
  maybe_result.sampe_rate #=> Integer
end  
```
and so forth?

From what I can see the approach we are taking is that each parser can only provide one nature of output, and we basically pick "the first" usable parsed result for each nature – is that correct? So given a situation with, say, a MOV parser and an MP4 parser, and with a CR2 and a TIF parser being separate entities, we will get either the MOV or the MP4 result _whichever parser did manage to produce output first_. This can work, but is intimately dependent on the definition order of the parsers. If the TIF parser is registered _last_ it might intercept a CR2 file and tag it as a TIF (which it is! totally!) and return the result, so the CR2 parser will not even get at it. This means that if we want to have these two parsers separate we need to hardcode _parser specificity,_ which I can't say I'm a fan of. Here is why: I feel like it can be quite possible that having multiple parsers correctly tag a file within the same nature (for example MOV and MP4) is exactly what we are going to encounter most. If that is going to be the case, how do we decide which parser should win?

## Interface

Making a parser something you can `call` with an input is awesome - I love the idea, because they are fundamentally functions with some state optionally encapsulated in the parser object. I think we should adopt this convention, maybe as a shortcut on class level even (to ensure instance variables are always reset and parsers do not have to be thread-safe) - this way we can avoid instantiating a new Parser if we know it isn't necessary for the parser to work:

```
def self.call(io)
  new.call(io) # if we really do a new instance
end
```

with which we can even do `SomeParser.(io)` and the simplest possible parser can be a Proc.

## Namespacing

See what's unfolding in #36. Given that is the case, is it wise to shoehorn every parser into a namespace and/or family of output? Is that necessary to register that parser under a single nature only? Especially given that with the discovery that a parser can be callable it might not even be a Module to begin with? I would say this actually makes things quite a bit harder to manage since the structure you propose, if implemented, for #36 specifically would mean having a separate copy of that parser for each filetype/nature which are very different. As well as ether sharing the parsed atom tree between parsers or having to re-parse this atom structure anew for each type.

## Mutex and parallelisation

There is a very specific reason why I've put a Mutex in there in the first place. Literally all of our background workers are threaded. We are also using threads in all of our services (we run them in Puma) so being thread-safe is very important. Since we allow parsers to register themselves - which is done so that other libraries or applications can add their own parsers as plugins - this registration procedure should be able to work even in a threaded context. Since array appends are only incidentally atomic (sometimes) on MRI they should be protected with a mutex, if you run this code on say jRuby you are likely to miss a parser registering itself since you will drop the array state during read+write. This is why registering a parser had to be made thread safe - to allow other apps to register parsers _with us._ Needless to say, having frozen constants with parser lists is also not inviting for that usage pattern.

## Parallel parsing

This *might* be an option in the future but we haven't pursued it right now - here's why. When we are parsing, it's mostly tight loops over IO output, which are not parallel on MRI - they are "between-the-lines-concurrent" rather. So we could do parsing in parallel, but if we decided to do so we wouldn't have gained that much performance. But we are making use of the fact that parsers use a shared page cache (see `Care`) so that if one parser happens to need one byte at offset 1234, the subsequent parsers will also have access to that byte, in memory. If we were to do this in a multithreaded way, we would need to either build out the library around a Fiber (like FastImage does), or we would need to have a thread-safe cache. Both are sane options but they are lower on the priority list - currently we make use of the fact that we can, in fact, parallelise when HTTP calls are executed. Patron releases the GIL, which allows us to preempt other Ruby threads during CURL execution, which gives us plenty of parallelism. But if you want to take a stab at making `Care` thread-safe we can transform the parsing pipeline into scatter-gather sooner - you would need a separate `IOWrapper` probably for each thread, with each `IOWrapper` using the same thread-safe instance of the Care page cache - a separate IOWrapper is needed because the IO object in Ruby is stateful so at the minimum you want each parser to maintain it's own `IO#pos`, or at least find the IO at the same position it has left it at every time :-)

## Return values

This is where I think this PR shines. IMO we should concentrate on making the _output_ we give to the caller more fine-grained - for example, not mixing audio attributes into an image result and vice versa. So if anything, maybe a hierarchy like the one proposed can be a better fit for the results we return?

## Style

Like I mentioned - I love the switch to `call` for the parser API. I do have doubts about `inject` - it has it's uses for sure, but a) it's usually easier to use `each_with_object` when building things like hash tables or arrays and b) why use it at all if what you are looking for is findamentally an `each` combined with an early return? I like the "break with result" idiom but I have never seen it before. Is an explicit early return not something used more often?

`each_with_object` vs. `inject` can be debated, but the way I see it is this: Ruby is _not_ very copy friendly, and attempts to force it to behave like a language with immutable core types can work out but with performance penalties. For example, a "Hash-collecting" `inject` allocates a new Hash if not careful, whereas an `each_with_object` is honest with stating "this is a language with mutable types and we are going to make use of it".

There is a number of smaller things that could be adjusted but there is a major useful thing emerging here that I don't want to spoil, so let's try to fledge that out first ;-)

@bermannoah your thoughts?

bermannoah:
Thanks much for making the changes - do feel free to create the style issue if you'd like! 

Regarding this PR: I basically just have an enthusiastic +1 of @julik's comments - I can understand the thinking behind placing each parser into a family, but I think for this library's specific use case it really limits our flexibility. I have an internal project that _does_ filter everything into media type "families" because we'll be doing different things based on the kind of file, not necessarily the specific file type - we can send all images to one processing service, for example. On the other hand for format_parser we want to be as specific as possible about the individual files so services like mine or others can make the decisions they need to make about each file.

That's a very roundabout preface to the question here: what should this actually return if a file can be two different kinds of things? What if we do something like reversing the hierarchy you've got here - so an MP4 can be in the "audio" family as well as the "video" family? Maybe somehow the caller can check "oh, it's an mp4 that's only in the audio family, I don't have to look for resolution data." _But_ then does a video-and-audio return as being both of those things or just video? 

(Maybe some of these questions are more 'semantics' than 'code' but as far as an interface for the user is concerned I think they're important considerations...)

tehAnswer:
Well, that was naive of me to think each parser would return one nature. With that in mind, At its core, dealing with multiple results has been translated to: 

# The Return value

One of the main pain points of the previous solution was the high amount of nils in the _model_ entity. Therefore, it's a good idea to have separate entities (`Audio`, `Image`, `Document` & `Video`) as-is regardless they're **_orphan_** (_get it? They have no more family. Hehe.  Not Funny._) now... do we need a wrapper object or can be it an array?

Each one has its advantages and disadvantages, and in my opinion, it finally comes down to one's taste. Maybe the best option is to have a wrapper object because it can be easily converted into an array and it can delegate into the latter to have available `Enumerable` capabilities.

```ruby
results = FormatParser.parse(io)  # => <#FormatParser: @results=[<#FormatParser::Image>]>
results.each { |result| print result }
``` 

# The Heuristic to stop the computation

This point boils down to two aspects: chameleonic formats and parsing performance. As you know better, the gem returns the first result it gets and, as indicated that's problematic. Not only there's the risk for the function of not being pure, it could also lull us into some weird "preference system" where parser A has to be called before parser B and whatnot, mostly regarding these tricky formats that extend other specification.

But what is the problem though? I think the problem is that the parsing has to be as cheap as possible. Maybe we could kill two birds with one stone via adding more syntax to the calls. For example, I could imagine something like:

```ruby
# Faraday style
parser = FormatParser.new do |config|
  config.natures = [:videos, :documents]  # formats we're interested in
  config.limit = 3                                          # number of results we want, default all.
  config.as = :array                                      # IDEA: type of return (e.g array, openstruct, wrapper)
end

parser.parse(io)
```

or with a chainable interface

```ruby
FormatParser.only(:videos, :documents).limit(3).as(Array).parse(io)
``` 

In these scenarios, the parsers would be statically tagged (I imagine a class method returning an array of symbols) that would be later used to filter parsers out, just to save some ms. I'm not 100% sure this would be more performant than what it's already here, but for sure will become handy as the number of formats grows. Nevertheless, that could be a smart way to stop the computation, just ask the user what it wants.


If this sounds promising, I can prepare another PR @bermannoah @julik.

🤡

bermannoah:
My initial thought is that part of the advantage of the current setup is that we don't need to know anything about the files beforehand. How do we determine whether or not we want to activate the `Foo` parser but not the `Bar` parser? 

One of the things I experimented with was a thing that would first determine filetype and then select a parser - with only the file as input - but for some filetypes that's an occasionally extremely inexact science. (MP3s, for example, can have a really _fun_ variety of header bytes.)

That being said though, I also have some misgivings about the... performantness of the gem if it tries everything before it returns a result. Maybe we could introduce some kind of "make_it_go_super_fast" option so that it immediately bails and returns once it has one result, but also a "give_me_all_the_things" option if you're fine with it taking the extra time? (We're talking fractions of `ms` here but that can add up. 😬)

bermannoah:
Maybe there's some kind of... parent... object that does a zippy pattern match of the file, picks the parsers, and runs from there, adopting our orphans so to speak? 😉 @julik has a library called magic_bytes that does this, but we opted not to use it (see #26) because we basically do the same work in a different way here.

tehAnswer:
@bermannoah Hey dude! Some comments on your comments.

> My initial thought is that part of the advantage of the current setup is that we don't need to know anything about the files beforehand.

But if you think it through, these chainable calls or configuration procs are meant to be all optional, so it'd work without them too. 🙃

> Maybe there's some kind of parent class/module/thing that does a zippy pattern match of the file, picks the parsers, and runs from there 

Ignoring what "zippy pattern" means, isn't it almost exactly the same of your previous experiments, where you tried to detect the file type and then pick a parser? Or the only change would be that this time we have multiple parsers?

> adopting our orphans so to speak?  :wink:

Hehe. Funny.

bermannoah:
@tehAnswer Ah hey, think I misunderstood. If the chainable calls / procs can be optional then I'm all for it! 😄 

As far as the pattern match thing is concerned - I think there's gotta be a way of writing it so that it doesn't fall into the same trap as the experimental system. What if we use it at the end of the parsing process and in ... reverse - so that it returns all relevant families for a filetype?

tehAnswer:
@bermannoah Are you suggesting some sort of multiple Chain of Responsibility pattern? Or in other words, once that a successful result was found, it also returns the next go-to parser, so it jumps straight ahead to such parser and so on? If so, it could be a great idea, but for sure we'll need to keep some state or breadcrumbs in order not to fall into an infinite loop. 

🌀

bermannoah:
Nothing that complicated - basically taking what you've suggested above and reversing it, so rather than requesting :audio and :video we return a file with :audio and :video families, or just :audio, or just :video. 

Here's a rough sketch of what such a return could look like:

```
file_info = FormatParser.parse(File.open("myvideo.mp4", "rb"))
file_info.file_nature           #=> [:video, :audio]
file_info.file_format           #=> :MP4
file_info.width_px              #=> 320
file_info.height_px             #=> 240
file_info.audio_codec           #=> :AAC
```

and so on and so on. If it's an MP4 with no video we can just give it a nature of :audio. This way the caller can adjust whatever it does with the response without having to figure out what, say, `width_pix = 0` means. :)

bermannoah:
I think I'm doing a bad job of explaining myself and that's on me, my apologies - I definitely don't want endless lists of nils or for the parsing to _depend on_ the natures, rather that the natures give a sense of what we can expect from the file - just as they are currently, where the nature is set when the FileInformation is created and this depends on a completed parse.

An array of matches or a wrapper sounds ideal! If you've got an implementation you like then please write it on up! :)

tehAnswer:
@bermannoah I'm closing this PR and preparing another one to tackle the same issue.

tehAnswer:
@bermannoah D'oh, I forgot I don't have rights. Can you close it for me?

💋

julik:
@tehAnswer awesome, thanks!

bermannoah:
@tehAnswer Awesome, looking forward to it! :)

PR #38, opened 2018-01-04
Add IO methods required by exifr

bermannoah:
EXIFR kindly requests the presence of `getbyte` and `readbyte` IO methods, and here they are.

### DIFF
--- a/lib/read_limiter.rb
+++ b/lib/read_limiter.rb
@@ -36,8 +36,19 @@ class FormatParser::ReadLimiter
 
     @io.read(n)
   end
+  
+  # A particular implementation of read meant to make reading JPEG EXIF data
+  # easier on remote reads, but if we want to override the defaults and 
+  # use it somewhere else we can. (EXIFR expects just plain ol' readbyte so
+  # we have to include the defaults)
+  def readbyte(num_bytes_to_read: 1, unpack_code: "C")
+    @io.read(num_bytes_to_read).unpack(unpack_code).first
+  end
+  
+  # EXIFR requires a getbyte method that does exactly the same thing as readbyte
 
   def getbyte
-    @io.read(1).unpack("C").first
+    readbyte
   end
+  
 end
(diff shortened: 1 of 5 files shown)


lib/read_limiter.rb
```diff
@@ -36,4 +36,18 @@ def read(n)
 
     @io.read(n)
   end
+  
+  # A particular implementation of read meant to make reading JPEG EXIF data
+  # easier on remote reads, but if we want to override the defaults and 
+  # use it somewhere else we can. (EXIFR expects just plain ol' readbyte so
+  # we have to include the defaults)
+  def readbyte(num_bytes_to_read: 1, unpack_code: "C")
```
julik: Uhm... no. Here is why. `IO#readbyte` is a standard Ruby method, and it actually accepts no arguments. What we have done so far is that we have allowed calls to a _subset_ of IO's methods, used in a _subset_ of ways. For example, we allow `IO#read(<Integer>)` but we do not allow `IO#read`. Neither do we allow `IO#read(<Integer>,<String>)` and other such forms, of which there are legion. However, in so doing, we are only _reducing_ the API surface that IO provides. This guarantees that everything _we_ need to call will work with that subset, and for our limiters/delegators we can make our implementations smaller. What you are proposing here is taking a builtin Ruby method and _extending_ it with some additional functionality the original `IO#readbyte` does not provide. If we do that, when we run some Parser against a simple IO/File - like we do in our tests - it will fail if the parser incidentally uses these extra arguments. If you want a method for `read_and_unpack` - and this is what this method does - could it work better when implemented in a specific parser that wants it?

This probably bears a bit more explanation. When setting up the IO stack for this library I deliberately reduced the subset of the IO API that we _require_ to be available. This serves two goals: it's possible to feed separate parsers StringIOs and File objects directly, _and_ it makes the proxies we have to make to limit reads, implement caching and so forth thinner. Having said that, IMO we should add something within the parsers themselves that will limit the methods that _may_ be called on the given IO/IO-ish object, so that we do not encounter these problems with local vs. remote resources. Imposing an "interface" if you will, a-la https://github.com/WeTransfer/zip_tricks/blob/master/lib/zip_tricks/streamer/writable.rb

PR #39, opened 2018-01-05
Introduce a restrictive IO subset wrapper

julik:
When setting up the IO stack for this library I deliberately reduced the subset of the IO API that we require to be available. This serves two goals: it's possible to feed separate parsers StringIOs and File objects directly, and it makes the proxies we have to make to limit reads, implement caching and so forth thinner. This change ensures we are only able to call the methods our IO wrappers actually support, regardless whether the parser is getting a File, a RemoteIO or one of our cache/limiters. This replaces #38 so that we can proceed to some more interesting business. For the examples of what wants `size` - ID3v1 tags are at the end of the file, as well as the ZIP central directory etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ file_info.width_px              #=> 320
 file_info.height_px             #=> 240
 file_info.orientation           #=> :top_left
 ```
+
 If nothing is detected, the result will be `nil`.
 
 ## Design rationale
--- a/lib/care.rb
+++ b/lib/care.rb
@@ -12,6 +12,10 @@ class Care
       @pos = 0
     end
 
+    def size
+      @io.size
+    end
+
     def seek(to)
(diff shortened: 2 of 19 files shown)


julik:
@bermannoah note that in one of the tests you are doing requests against WEBRick but you do not have guarantees it will be running, which will make the test suite hang - since tests are in random order, WEBRick will only boot before we start with the `Fetching data from HTTP remotes` spec. This introduces a transient test failure with a loooooong timeout, if we want to do this we should either mock the remote fetching stuff in the EXIFR parser spec, remove those specs - since now the IO is contstrained anyway, or start WEBRick for the entire suite, regardless. As a matter of fact...

julik:
oops - disregard, this stuff is in the right spec already ❤️

Issue #31, opened 2017-12-27
WAV parser


julik:
for WAV sound files, providing information similar to the AIFF parser

duplamatyi:
@julik I would like to start working on the WAV parser if possible. As WAV files sometimes might contain compression I would propose in a first step to implement the correct prying of the RIFF chunk, fmt chunk and the data chunk for WAV files where the format tag of the fmt chunk is 1(meaning no compression). Later maybe I could also include the case of WAV files with compression in my PR but maybe that could be a second PR as a future enhancement.

See:
https://web.archive.org/web/20080113195252/http://www.borg.com/~jglatt/tech/wave.htm
http://soundfile.sapp.org/doc/WaveFormat/

julik:
@duplamatyi 👍  go ahead

duplamatyi:
@julik Created [#PR 42](https://github.com/WeTransfer/format_parser/pull/42). After all I addressed both the standard PCM format and the non PCM format cases. The approach is really similar but slightly differs from the aiff_parser. Please let me know what you think.

julik:
Aand we got one ❤️

PR #42, opened 2018-01-07
WAV parser

duplamatyi:
Parsing of WAVE files. (see: [issue 31](https://github.com/WeTransfer/format_parser/issues/31))

According to the WAVE format specifications WAVE files can be broken down into 2 categories namely files where the 'fmt ' format chunk specifies PCM(format code 1) format and everything else. This PR deals with both scenarios.

In case of the PCM audio format files the 'fmt ' chunk and the size of the 'data' chunk need to be parsed, while in the case of non-PCM format files the 'fmt ' chunk and the sample length from the 'fact' chunk need to be parsed in order to calculate the duration in frames and seconds.

A major difference to the aiff_parser is that here I chose not to whitelist the most common chunks as in the case of WAVE files there are too many. Instead the parser will ignore any other chunk than the 'fmt ', 'data' and 'fact' chunks.

Also note that the current implementation is strict in the sense that:
* it expects the 'fmt ' chunk to precede the 'data' chunk
* it does skip the 'data' chunk if the audio format in the 'fmt ' chunk is not 1 and vice versa

As a result, there are some files which do not get detected as WAVE files even though most other software would detect them as WAVE files.

Besides that I tried to comment the parser code where I felt it's needed.


lib/parsers/wav_parser.rb
```diff
@@ -0,0 +1,102 @@
...
+    # chunks. In the latter case the order fo appearence of the chunks is
+    # arbitrary.
+    fmt_processed = false
+    fact_processed = false
+    fmt_data = {}
+    total_sample_frames = 0
+    loop do
+      chunk_type, chunk_size = safe_read(io, 8).unpack('a4l')
+      case chunk_type
+      when 'fmt '
```
bermannoah: Can we have a comment or an escape character here so it's clear that the space is deliberate?

spec/parsers/wav_parser_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'spec_helper'
+
+describe FormatParser::WAVParser do
+  # Fixtures prefixed with c_ are considered canonical
+  # while fixtures prefixed with d_ deviate from the standard.
+  Dir.glob(fixtures_dir + '/WAV/c_*.*').each do |wav_path|
```
bermannoah: Are these files creative commons / public domain licensed / do we have permission to use them? If not let's swap 'em out for ones we're able to use. I can spin up a few files for you if you'd like, just let me know. :)

lib/parsers/wav_parser.rb
duplamatyi: I agree, it's a bit strange to have a space character in the ID of one of the most basic chunk types. I added a comment to draw the eye.

spec/parsers/wav_parser_spec.rb
duplamatyi: @bermannoah Sorry for this licensing mess.

Here are the websites of origin and licenses of the files:
* c_11k16bitpcm.wav, c_8kmp316.wav are from [Wikipedia WAV article](https://en.wikipedia.org/wiki/WAV#Comparison_of_coding_schemes) and their license don't seem to be explicitly mentioned anywhere. They've been since 22 June 2007 on Wikipedia. 
* c_39064__alienbomb__atmo-truck.wav is from [freesound](https://freesound.org/people/alienbomb/sounds/39064/) and it has CC0 license.
* c_M1F1-Alaw-AFsp.wav, d_6_Channel_ID.wav are from a [McGill Engineering site](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/Samples.html) and their license is not clear.

Let me know if you consider that I should remove the files except the one licensed under CC0. And if you want to provide me some files that's also welcome. Otherwise I will make some time tomorrow night for searching/creating some free to use files. I guess [this](https://en.wikipedia.org/wiki/List_of_free_content_licenses#For_other_content_types,_including_audio_and_video) is a good list of acceptable licenses.

lib/parsers/wav_parser.rb
```diff
@@ -28,7 +28,7 @@ def information_from_io(io)
     loop do
       chunk_type, chunk_size = safe_read(io, 8).unpack('a4l')
       case chunk_type
-      when 'fmt '
+      when 'fmt ' # watch out: the chunk ID of the format chunk ends with a space
```
bermannoah: <3

spec/parsers/wav_parser_spec.rb
bermannoah: No worries. Thanks much! We do need the files to be able to be licensed/released under the MIT license. A lot of our existing spec files are from other open source (MIT licensed) repos, so you could see what else is out there -- or I can make a few lil files later today, just let me know. :)

duplamatyi: I really hoped that CC license was compatible with this project. But in this case I will create some files later today. Probably it's quicker than finding MIT licensed WAV files.

julik: @bermannoah Would that help if we made a list of fixtures we use and note where they come from and their licenses? I am almost certain CC with attribution is fine, and in case of a claim we will be able to quickly remove those assets.

bermannoah: Yes! I think so. I added a few to the readme a while back - but I wasn't sure if we could use CC under MIT. I guess if it's CC0 then we're good, or CC with attribution. (But then does our MIT license mean that future users can use them without attribution?)

julik: I'd say let's add a list of all the fixtures including their provenance and if there is a cease-and-desist against any of them we pull them on request. The gems we deliver do not have these files anyway since they get stripped during the gem build.

bermannoah: @julik Sounds good. And so then for @duplamatyi -- if he documents the sources/licenses of the existing files then I guess we're good? :)

julik:
@duplamatyi There is a fixture sources section in the README, can you add your files (with URLs to the sources) to that section please? Once that is done I'll merge the PR, and thank you for your work ❤️

julik:
👏

duplamatyi:
@julik and @bermannoah Thanks for accepting my work. Really happy that I could help.

lib/matyi.rb
```diff
@@ -0,0 +1,59 @@
+require_relative 'format_parser'
```
bermannoah: @duplamatyi mind removing this file? ;)

julik:
oh woops, trigger happy

lib/matyi.rb
bermannoah: ah nvm I've got it

duplamatyi: @bermannoah of course I had to do something silly in the last 100 meters... thanks for noticing!

bermannoah: :)

duplamatyi:
@julik thanks for removing it

bermannoah:
Pushed v0.1.4 to rubygems and added WAV to our list o' things we can parse in the readme.

PR #43, opened 2018-01-07
MP3 parser + ID3v2 + ID3v1

julik:
Implements a parser for ID3v1, a basic one for ID3v2 and a parser for data frames (CBR+VBR). Only recovers the very basic information.


julik:
@bermannoah This should provide us with the basics. Can be improved and treated ad-infinitum of course, but as far as duration and stereo/mono goes we are good.

PR #47, opened 2018-01-11
Enabling multiple results

tehAnswer:
For a longer description, see #21

Hello again @julik and @bermannoah. In the last episode of "_How to Treat Multiple Values_" we came to the understanding that parsers couldn't be segregated into _families_, that we needed a way to return more than one value (archived now via returning an array) and that a compromise between the response completion and performance should be found. This pull request suggests a way to tackle the last two points, with an emphasis in the last one.

During our previous discussion, I came up with the idea of using some optional parameters to narrow down the parsing. I suggested two alternatives for the interface: a chainable interface and configuration proc. I ended up implementing the configuration proc but not the chainable interface. I also added support for keywords arguments so as using a proc is not enforced.

I decided to ditch the chainable interface because it was too verbose and fancy for what we wanted. Also, it didn't quite offer a good fit for this use case and the implementation would pollute the object space with one or more instance vars per step of the chain.

To validate the user input, I decided to add a small dependency which you're also familiar with it. `dry-structs`. Yeah. `dry-structs`. @bermannoah I am aware you added this dependency and then quickly removed it. Well, I suggest just to add it again. I just did it to enforce some type checks for these options that the user of the gem can now pass. Why? Never trust user input and... don't reinvent the wheel. I was tempted also to make the entities of our model `Dry::Structs` as well, but I finally decided not to just because @bermannoah walked away from it in the past. Was it because of horrible error messages though?

Anyway, in other order of business, I added the features you liked in the previous pull request. So that means that `FileInformation` is now divided in `Format::Image` and `Format::Audio`, that `information_from_io` has been renamed to `call` and other minor renames (`file_nature` => `nature`, `file_type` => `format`). Finally, I also made some minor improvements in the code, completely unrelated to the purpose of this pull request just for... well, I couldn't contain myself. 😅

## TL; DR

1. Now `FormatParser.parse(...)` returns an array of matches.
2. Now `FormatParser.parse(...)` can receive options to narrow down the parsing with a proc or keyword arguments:

```ruby
matches = FormatParse.parse(file) do |config|
  config.natures = [:audio]
  config.formats = [:mp3]
  config.limit = 1
end

# same as
matches = FormatParser.parse(file, natures: [:audio], formats: [:mp3], limit: 1)
```

3. `FileInformation` has been divided into multiple classes now.
4. Improvements in method and attributes naming.


julik:
This is very interesting. I think we walked away from dry structs for the following reasons:

* Nillable types were a pain
* Indeed horrible error messages
* The file information that a parser can provide would invariable have some - or even most - of the returned properties not set, which means we would have to explicitly set them to `nil` at the very least. Which, given the number of properties, was insane. We also need to pass `intrinsics` now which is an arbitrarily-typed property basically, and dry-rb doesn't like anything arbitrary - they think that writing Haskell in Ruby is a good idea.

I will look at this PR in detail tonight or tomorrow, lots of interesting concepts here and I rather like it I must say 💪

bermannoah:
+1 that this is super interesting. Will give it a more detailed read-through tomorrow but I'm digging it. Really like your approach with dry-structs -- I abandoned it for all the reasons both of you mention and also especially because it makes optional-but-specific-properties (if that makes sense) infuriatingly difficult to use, but with your implementation that's not an issue! 🎉 Huzzah.

lib/format_parser.rb
```diff
@@ -30,24 +39,30 @@ def self.parse_http(url)
     parse(cached_io)
   end
 
-  def self.parse(io)
+  def self.parse(io, **opts, &proc)
```
julik: To be honest I'm not convinced of the block+opts approach. All of our code is Ruby2+ so we can use keyword arguments at will. Keyword arguments also give us a sane interface to specify default values. For example, we can use magic Symbols to signify a default with which you want all the natures:

```
def self.parse(io, accept_natures: :all, ...)
```

Should we want to use something as an argument for externally specified configuration everywhere, we can use any object that responds to `to_hash` and pass it in as the replacement for keyword arguments:

```
FormatParser.parse(io, **FORMAT_PARSER_CONFIG)
```

I.e. I do get the value of block configuration, but I don't see it in this use pattern. Moreover, we should be distrustful of the _file contents_, but not of the way how someone is using the library. We trust the user to make the right choices and if there is an incorrect file nature passed we can capture it at runtime - this will be a nature no parser supports.

A very valid example of block configuration is when it is divorced from the call site. For example:

```
SomeLibrary.configure |c|
  c.use_evil_corporate_http_proxy_from_microsoft = true
end

#... in another module that does something useful
SomeLibrary.perform_dispatch(name, title) # will indeed use the draconian proxy
```

but we are not doing it here and I don't think we might want to. @bermannoah how is the ATC usage pattern looking so far with regards to this?

lib/format_parser.rb
```diff
@@ -30,24 +39,30 @@ def self.parse_http(url)
...
     # If the cache is preconfigured do not apply an extra layer. It is going
     # to be preconfigured when using parse_http.
     io = Care::IOWrapper.new(io) unless io.is_a?(Care::IOWrapper)
+    results = []
 
     # Always instantiate parsers fresh for each input, since they might
     # contain instance variables which otherwise would have to be reset
     # between invocations, and would complicate threading situations
-    parsers = @parsers.map(&:new)
+    parsers = @parsers.select do |p|
```
julik: I think this block could use more lines because it does two things - it collects parsers we are willing to try and it instantiates them. That's a bit stuffed for one line. Moreover - it does selections on two criteria. Also either of these criteria can be set, while the other remains at `:all`, so it could be useful to do these selects in steps.

lib/format_parser.rb
```diff
@@ -30,24 +39,30 @@ def self.parse_http(url)
...
+      !(p.natures & config.natures).empty? && !(p.formats & config.formats).empty?
+    end.map(&:new)
```
julik: Do we really need to fence off the question methods about what parsers support in their class method namespace? Isn't it simpler to query parsers after instantiating them? We will create throwaway instances while doing so, but this will prevent too much bookkeeping within the parsers - of which we have many.

lib/format_parser.rb
```diff
@@ -30,24 +39,30 @@ def self.parse_http(url)
...
       io.seek(0)
       # Limit how many operations the parser can perform
       limited_io = ReadLimiter.new(io, max_bytes: 512*1024, max_reads: 64*1024, max_seeks: 64*1024)
       begin
-        if info = parser.information_from_io(limited_io)
-          return info
+        if info = parser.call(limited_io)
+          results << info
+          # Return early if the limit was hit.
+          return results if config.limit == results.length
```
julik: Can there be a case when we want something else than "only the first one" and "all of the ones matched"?

lib/format_parser.rb
```diff
@@ -58,7 +73,22 @@ def self.parse(io)
         # and examine the file more closely.
       end
     end
-    nil # Nothing matched
+    # Return the array of results if something matched or nil otherwise.
+    results.empty? ? nil : results
+  end
+
+  def self.parse_config(**opts, &proc)
```
julik: OpenStruct is known to be slow, besides - see my remark up the code about overinflating this interface. I think just having kwargs explicitly declared in the method signature with sane defaults will do.

lib/image.rb
```diff
@@ -1,13 +1,14 @@
 module FormatParser
-  class FileInformation
+  class Image
```
julik: This is one of the most valuable parts of this PR in my opinion 👍

lib/image.rb
```diff
@@ -67,8 +52,8 @@ def initialize(**attributes)
       attributes.map { |(k, v)| public_send("#{k}=", v) }
     end
 
-    def self.image(**kwargs)
-      new(file_nature: :image, **kwargs)
+    def nature
```
julik: Why don't we forego the constant and just return a symbol here? After all, Symbols are reused in Ruby. Besides, there is an `attr_accessor` which will work like this (if used at all):

```
image_info.nature = :powerpoint # will totally succeed
image_info.nature #=> :image
```

So we should have either an accessor, or a hardcoded reader - but not both.

lib/parse_config.rb
```diff
@@ -0,0 +1,10 @@
+module FormatParser
+  class ParseConfig < Dry::Struct
```
julik: Given that this is _a ton_ of infrastructure just to validate some keyword arguments, do we have a compelling reason to use Dry::Struct? We wanted to use it to make it harder for parsers to return nonsensical information about misparsed files (for example, some image parser returning an image witdh of `negative two million` or some such - which can happen if, for example, you unpack an unsigned int as if it were signed - but not for validating programmer input.

lib/parsers/aiff_parser.rb
```diff
@@ -18,8 +18,20 @@ class FormatParser::AIFFParser
...
 
-  def information_from_io(io)
+  NATURES = [:audio].freeze
+  FORMATS = [:aiff].freeze
+
+  def self.natures
+    NATURES
+  end
+
+  def self.formats
```
julik: See above. Why don't we let the parser objects tell us which formats they are willing to accept after they are instantiated? Also, who not just return a symbol - i.e. why do we need constants - that we also have to get right in every parser? Moreover - what if we let every parser answer a question as opposed to returning a hardcoded list:

```
def will_emit_format?(some_format)
  ...
end
```

Then you can keep any allocations you might have to do for this within the method. It will also simplify your `select()` in the main `parse()` method  (`accepted_natures.any?{|nature| parser.will_emit?(nature) }`).

Computationally it will be slightly more overhead but this is peanuts compared to the actual grunt work parsers have to do.

lib/parsers/aiff_parser.rb
```diff
@@ -18,8 +18,20 @@ class FormatParser::AIFFParser
...
+    FORMATS
+  end
+
+  def call(io)
     io = FormatParser::IOConstraint.new(io)
+    io.seek(0)
```
julik: These might be redundant because `parse()` already seeks to 0 for every parser in succession. The constrained IO is used in every parser deliberately because parsers usually get tested with just `File` objects and we want to forcibly reduce their API surface.

lib/parsers/dpx_parser.rb
```diff
@@ -124,7 +124,18 @@ class FormatParser::DPXParser
   LE_MAGIC = BE_MAGIC.reverse
   HEADER_SIZE = SIZEOF[DPX_INFO] # Does not include the initial 4 bytes
 
-  def information_from_io(io)
+  NATURES = [:image].freeze
```
julik: See comments for the AIFF parser, comments are mostly identical.

lib/format_parser.rb
```diff
@@ -58,7 +73,22 @@ def self.parse(io)
         # and examine the file more closely.
       end
     end
-    nil # Nothing matched
+    # Return the array of results if something matched or nil otherwise.
+    results.empty? ? nil : results
```
julik: If we go through our parsers and return multiple results, why not turn this into a `map`?

lib/format_parser.rb
```diff
@@ -58,7 +73,22 @@ def self.parse(io)
         # and examine the file more closely.
       end
     end
-    nil # Nothing matched
+    # Return the array of results if something matched or nil otherwise.
```
julik: Returning an Array is a good thing here indeed, the simplest conceptually and has a predictable API. It _might_ have merit to introduce some flag (see up the code) to differentiate between "I want the first result whatever it might be" and "I want all the results and I want to make the decisions myself", which could indeed use either an early return within a map (see below)

lib/parsers/tiff_parser.rb
```diff
@@ -1,34 +1,36 @@
...
     scanner = FormatParser::EXIFParser.new(:tiff, io)
     scanner.scan_image_exif
-    if scanner.orientation
-      FormatParser::FileInformation.image(
-        file_type: :tif,
+    return FormatParser::Image.new(
+        format: :tif,
         width_px: w,
         height_px: h,
         orientation: scanner.orientation
```
julik: Good call to remove the conditional, but I would appreciate a comment along the lines of `might be nil if no EXIF metadata was found` or something to that end

spec/format_parser_spec.rb
```diff
@@ -20,4 +20,47 @@
       end
     end
   end
+
+  describe 'multiple values return' do
+    let(:blob) { StringIO.new(Random.new.bytes(512 * 1024)) }
+    let(:audio) { FormatParser::Audio.new(format: :aiff, num_audio_channels: 1) }
+    let(:image) { FormatParser::Image.new(format: :dpx, width_px: 1, height_px: 1) }
+
+    context '#parse called without any indication' do
```
julik: "indication"?

lib/format_parser.rb
```diff
@@ -14,6 +18,11 @@ def self.register_parser_constructor(object_responding_to_new)
     PARSER_MUX.synchronize do
       @parsers ||= []
       @parsers << object_responding_to_new
+      # Gathering natures from parsers.
```
julik: See below. We are indeed achieving a speedup by having parser natures cached here, but unless I see strong indications that this speedup is extremely significant compared to the time it takes to actually parse things this looks like overcomplicating things - instead of just asking an already instantiated parser "will you be emitting images?" and discarding it if we don't need it. There is GC overhead to everything but in this case it should not be that big of an issue, at least initially.

lib/format_parser.rb
bermannoah: Agree on swapping back to keyword args. Since ATC uses FormatParser once per file parse job, it would end up using `accept_natures: :all` anyway.

lib/format_parser.rb
bermannoah: Also there's quite a bit happening here - if we _do_ need something like this I think it could be made more explicit. (Maybe just my personal styling pref though.)

lib/format_parser.rb
bermannoah: +1 for what Julik said, and also - maybe a silly question but would it make sense to return an empty array here if nothing matches?

lib/image.rb
bermannoah: 🙇 🙇 agreed. Also matches suuuuper well with the system I'll be using this in internally, which will let me refactor out some complexity.

lib/parse_config.rb
bermannoah: I do like this implementation of Dry::Struct for keeping input consistent, but I do agree that if we're going to add a dependency like this we ought to use it to validate output as well.

lib/format_parser.rb
tehAnswer: In my design, we need to gather all the natures from parsers to provide default parsing options. The whole motto of this PR is to, like you said, trust that the user will make the right choices. Therefore, in case of our user will not provide any _indication_ (😅), it's assumed that she rather have all the results available. The code as-is, to not to filter out any, requires knowing what are the default values. Because of the whole mutex thingy, I thought to make it extensible rather than using a hardcoded array.

However, in case you don't like it, we can use a magic constant (`:*`) to avoid this gathering, but that's gonna make the code more brittle imho.

lib/format_parser.rb
tehAnswer: Hmmm, it's true that the block code looks superfluous right now. I think it would become handier when the number of options would wild up. I'll remove and kill dry-db dependency alongside.

lib/format_parser.rb
tehAnswer: I don't agree. It's doing one thing only based on two criteria. Both can be set at the same time, it's not exclusive: nobody is stopping you to do something like:

```ruby
FormatParser.parse(f, natures: [:audio], formats: [:mp3])
```

lib/format_parser.rb
tehAnswer: Both cases are now supported. 

```ruby
FormatParser.parse(f, limit: 1)
FormatParser.parse(f) # it trust that a parser only returns one value
```

lib/format_parser.rb
tehAnswer: +2, I wanted to modify it but I wasn't sure if there was some code depending on a nil to trigger some errors or whatnot!

lib/format_parser.rb
tehAnswer: Do you any advantage in returning the following map versus the current array?

```ruby
{
  mp4: <FormatParser::Video format=:mp4 ... >,
  mp3: <FormatParser::Audio format=:mp3 ... >, 
  mov: <FormatParser::Video format=:mov ... >
}
```

For sure there are faster lookups for formats, but does that make sense when we're planning to add a filter based upon formats? Also, in case that the keys were different, let's say based on natures, wouldn't it be overdoing? The previous principle applies.

I see some overlapping, therefore I'd pass on this one.

lib/image.rb
tehAnswer: The attribute_accesor might a leftover from FileInformation, it shouldn't be there. Just a hardcoded reader is what we need.

README.md
```diff
@@ -15,15 +15,15 @@ and [dimensions,](https://github.com/sstephenson/dimensions) borrowing from them
...
-file_info.width_px              #=> 320
-file_info.height_px             #=> 240
-file_info.orientation           #=> :top_left
+matches = FormatParser.parse(File.open("myimage.jpg", "rb"))
+matches.first.nature        #=> :image
+matches.first.format        #=> :jpg
+matches.first.width_px      #=> 320
+matches.first.height_px     #=> 240
+matches.first.orientation   #=> :top_left
 ```
```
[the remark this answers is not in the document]
tehAnswer: `matches` is the name of the variable and `first` it's already an alias I think. 😉

spec/format_parser_spec.rb
tehAnswer: Will rename it to option. 

🤔

lib/parsers/aiff_parser.rb
tehAnswer: Is there any reason why you want this at the instance level and not in the class level? However, I like the idea of a question method. That can make the code more readable even though it adds some degree of code duplication, but nothing crazy that can't be assumed.

README.md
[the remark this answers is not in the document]
bermannoah: Right right right I mean if we call `matches.format` it' could return just `:jpg`. Don't know if it's necessarily necessary but it'd be a 🆒 convenience to have.

lib/format_parser.rb
julik: I don't think I mentioned "exclusive". What I meant was that you can do this:

`parse(f, natures: [:audio])`

as well as

`parse(f, formats: [:png, :tif])`

as well as both. Each of these keyword arguments can be turned into a separate pass of the `.select`, which, while not as optimized, will be more readable.

lib/format_parser.rb
julik: That's not what I meant. We are currently using an `.each` do go over the parsers. Since we now preserve multiple return values, we could turn the `.each` into a `.map` and then `.compact` the resulting Array

lib/parsers/aiff_parser.rb
julik: The reason is that it's easier to write a parser implementation then - the mental model of a parser is going to be simpler, you won't have to remember where you need class methods and where you need instance methods.

README.md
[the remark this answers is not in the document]
tehAnswer: Done. 💯

lib/parsers/aiff_parser.rb
tehAnswer: Done.

lib/format_parser.rb
tehAnswer: Done!

julik:
Awesome!

lib/format_parser.rb
```diff
@@ -30,25 +41,32 @@ def self.parse_http(url)
...
 
+    # How many results has the user asked for? Used to determinate whether an array
+    # is returned or not.
+    amount = case returns
+             when :all
+               @parsers.count
+             when :one
+               1
+             else
+               throw ArgumentError.new(":returns does not match any supported mode (:all, :first)")
```
bermannoah: Does it expect `:first` or `:one` here? && could you add a note about this to the readme? :)

bermannoah:
Bumped to 0.2.0, added a lil post-install alert, pushed to rubygems! 🎉

PR #49, opened 2018-01-13
FDX parser

bermannoah:
Closes #49

### DIFF
--- a/lib/parsers/fdx_parser.rb
+++ b/lib/parsers/fdx_parser.rb
@@ -1,0 +1,31 @@
+class FormatParser::FDXParser
+  include FormatParser::IOUtils
+
+  def information_from_io(io)
+    if xml_check(io)
+      file_and_document_type = safe_read(io, 100)
+      is_it_final_draft?(file_and_document_type)
+    else
+      return # Bail if it's not even XML
+    end
+
+    file_info = FormatParser::FileInformation.new(
+      file_nature: :document,
+      file_type: :fdx
+    )
+  end
+
+  def xml_check(io)
+    xml_check = safe_read(io, 5)
+    xml_check == "<?xml" ? true : false
+  end
+
+  def is_it_final_draft?(file_and_document_type)
+    if file_and_document_type.include?("FinalDraft")
+      true
+    else
+      return
+    end
+  end
+
+end
--- a/lib/parsers/fdx_parser.rb
+++ b/lib/parsers/fdx_parser.rb
@@ -4,14 +4,15 @@ class FormatParser::FDXParser
   def information_from_io(io)
     if xml_check(io)
       file_and_document_type = safe_read(io, 100)
-      is_it_final_draft?(file_and_document_type)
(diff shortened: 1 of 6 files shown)


lib/parsers/fdx_parser.rb
```diff
@@ -0,0 +1,33 @@
...
+  include FormatParser::IOUtils
+
+  def information_from_io(io)
+    if xml_check(io)
+      file_and_document_type = safe_read(io, 100)
+      file_type, document_type = check_for_document_type(file_and_document_type)
+    else
+      return # Bail if it's not even XML
+    end
+
```
julik: Youa re having a bit of a gotofail here - if `check_for_document_type` returns a nil, you are going to return a `:document` FileInformation with a `nil` file_type, in practice you will be capturing _any_ XML document. Including, say, MusicXML and SVG.

lib/parsers/fdx_parser.rb
```diff
@@ -0,0 +1,33 @@
...
+  def check_for_document_type(file_and_document_type)
+    sanitized_data = file_and_document_type.downcase
+    if sanitized_data.include?("finaldraft") && sanitized_data.include?("script") 
+      return :fdx, :script
+    else
+      return
+    end
+  end
+
+end
```
julik: Newline

spec/parsers/fdx_parser_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    Dir.glob(fixtures_dir + '/*.fdx').each do |fdx_path|
+      it "is able to parse #{File.basename(fdx_path)}" do
+        parsed = subject.information_from_io(File.open(fdx_path, 'rb'))
+        expect(parsed).not_to be_nil
+        expect(parsed.file_nature).to eq(:document)
+        expect(parsed.file_type).to eq(:fdx)
+        expect(parsed.document_type).to eq(:script)
+      end
+    end
+  end
```
julik: You could use a test here that you are not intercepting other XML documents by accident (right now you do).

bermannoah: 👍

bermannoah:
Unfortunately but not unsurprisingly, the FDX files don't contain a "total number of pages" property. They _do_ contain a list of system voices to potentially use for each character, but no TotalPages info. This probably explains something but I can't think what.

julik:
> This probably explains something but I can't think what.

That one is easy - that they want you to buy FD to be able to compute the number of screen minutes, what else.

bermannoah:
"It's a feature, not a bug"

lib/parsers/fdx_parser.rb
```diff
@@ -0,0 +1,31 @@
...
+    sanitized_data = file_and_document_type.downcase
+    if sanitized_data.include?("finaldraft") && sanitized_data.include?("script") 
+      return :fdx, :script
+    else
+      return
+    end
+  end
+
+  FormatParser.register_parser_constructor self
+end
```
julik: Still a newline

PR #55, opened 2018-01-16
Rubocop integration

linkyndy:


linkyndy:
@julik @bermannoah, I made a few (more) tweaks to our style gem, fixed most offenses and here is the outcome. There are still 7 offenses, please check Travis and tell me how should we proceed with them (don't know exactly the logic and reasoning for those lines).

PR #61, opened 2018-01-22
Reverse API changes to support :first as default and add opts to parse_http

### LINKED ISSUE #60: Remote parser does not have natures/formats/returns options

bermannoah:
Closes #60

### DIFF
--- a/spec/format_parser_spec.rb
+++ b/spec/format_parser_spec.rb
@@ -37,7 +37,7 @@ describe FormatParser do
       it { is_expected.to include(image) }
       it { is_expected.to include(audio) }
     end
-\
+
     context '#parse called without hash options' do
       before do
         expect_any_instance_of(FormatParser::DPXParser).to receive(:call).and_return(image)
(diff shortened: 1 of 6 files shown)


bermannoah:
@julik switched to using `:first` as default rather than always returning an array. Feel like it makes sense to have "return all the things" be optional rather than enforced?

spec/remote_fetching_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 require 'webrick'
 
-describe 'Fetching data from HTTP remotes' do
+describe 'Fetching data from HTTP remotes' do
```
julik: Stray space

spec/format_parser_spec.rb
```diff
@@ -26,24 +26,24 @@
...
 
-      subject { FormatParser.parse(blob) }
+      subject { FormatParser.parse(blob, results: :all) }
 
       it { is_expected.to include(image) }
       it { is_expected.to include(audio) }
     end
-
-    context '#parse called with hash options' do
+\
```
julik: Stray slash

julik:
@bermannoah There is also a fuzzing error in the MP3 parser - it can be reproduced on this branch, see https://travis-ci.org/WeTransfer/format_parser/jobs/332196495

bermannoah:
@julik this is a weird one. Think it's coming from the read method in IOWrapper. Pushed a fix, let me know. :)

PR #62, opened 2018-01-24
Use register_parser for natures/fmts

julik:
So @tehAnswer had a point here - we can simplify the code
quite a bit if we preselect parsers matching the natures/formats
we expect. However, there is an additional benefit. Currently, we
effectively require something registered as a parser to support the
following API:

  #natures
  #formats
  #call(io)
  .new

In the new setup we bring this down to one method:

  #call(io)

How do we do this? Well, first we move the two
methods you would have to manually define into the
call to `register_parser`. When we register a parser,
we pass either a callable or a class, and the module
figures out whether it has to call #new on the thing
given to it. This spares us from having  to have `.new`
defined - and thus spares us having to have parsers
be a Class in the first place. We can make a parser that
is a Proc instead, as long as we register it having
the right natures/formats.

Second, we supply the natures and the formats as the arguments
to the `register_parser` method - we tell FormatParser that
when someone wants to have anything of the kind of `:image`,
our parser is one that can provide that. We don't need to
supply that data from the parser object necessarily, and when
I asked to have this I was actually mistaken - because any
interface beyond just something callable is in this case
redundant.

So now, when we do the parser registration, we can do this

   FormatParser.register_parser self, natures: :image, formats: :jpg

and the module will figure out that it has to call .new on us
if we respond to it, and if we don't - the we will be considered
callable as is


README.md
```diff
@@ -43,45 +43,94 @@ FormatParser.parse(File.open("myimage", "rb"), natures: [:video, :image], format
...
 
-1) Instances of the new parser class needs to respond to a `call` method which takes one IO object as an argument and returns some metadata information about its corresponding file or nil otherwise.
-2) Instances of the new parser class needs to respond `natures` and `formats` accessor methods, both returning an array of symbols. A simple DSL is provided to avoid writing those accessors.
-3) The class needs to register itself as a parser.
+The IO will at the minimum support the subset of the IO API defined in `ConstrainedIO`
 
+Strictly, a parser should be one of the two things:
+
+1) An object that can be `call()`-ed itself, with an argument that conforms to `ConstrainedIO`
+2) An object that responsds to `new` and returns something that can be `call()`-ed with the same convention.
```
tehAnswer: Typo in the fourth word!

README.md
```diff
@@ -43,45 +43,94 @@ FormatParser.parse(File.open("myimage", "rb"), natures: [:video, :image], format
...
+  )
+}
+
+# Register the parser with the module, so that it will be applied to any
+# document given to `FormatParser.parse()`. The supported natures are currently
+#      - :audio
+#      - :document
+#      - :image
+#      - :video
+FormatParser.register_parser MyParser, natures: :image, formats: :bmp
```
tehAnswer: This is great! Good one.

README.md
```diff
@@ -43,45 +43,94 @@ FormatParser.parse(File.open("myimage", "rb"), natures: [:video, :image], format
...
+```
+
+If you are using a class, this is the skeleton to use:
+
+```ruby
+class MyParser
+  def call(io)
+    # Make sure you are only using the right subset of IO methods
+    io = FormatParser::ConstrainedIO.new(io)
```
tehAnswer: Hm. Hmmm.

I think this is a code smell. To reassign an argument makes me think whether the parameter passing in the method invocation was imprecise (maybe a ConstrainedIO has to be enforced/assumed) or that you need a need var. I honestly dislike to modify arguments or reassigning variables. Just my two cents though. 🙃

lib/format_parser.rb
```diff
@@ -8,21 +8,34 @@ module FormatParser
...
-      @natures ||= Set.new
-      # NOTE: merge method for sets modify the instance.
-      @natures.merge(parser.natures)
-      @formats ||= Set.new
-      @formats.merge(parser.formats)
+      @parsers ||= Set.new
+      @parsers << callable_or_responding_to_new
+      @natures_per_parser ||= {}
+      parser_provided_natures.each do |provided_nature|
+        @natures_per_parser[provided_nature] ||= Set.new
```
tehAnswer: Should be this call parsers_per_nature? After all, it's an inverted index for parsers given a nature. Same goes for format.

lib/format_parser.rb
```diff
@@ -8,21 +8,34 @@ module FormatParser
...
+        @formats_per_parser[provided_format] << callable_or_responding_to_new
+      end
+    end
+  end
+
+  def self.deregister_parser(callable_or_responding_to_new)
+    # Used only in tests
+    PARSER_MUX.synchronize do
+      (@parsers || []).delete(callable_or_responding_to_new)
+      (@natures_per_parser || {}).values.map { |e| e.delete(callable_or_responding_to_new) }
```
tehAnswer: I'd use `&` operator to be nil-safe.

lib/format_parser.rb
```diff
@@ -78,16 +93,34 @@ def self.parse(io, natures: @natures.to_a, formats: @formats.to_a, results: :fir
...
+    factories = fitting_by_natures & fitting_by_formats
+
+    if factories.empty?
+      raise ArgumentError, "No parsers provide both natures #{desired_natures.inspect} and formats #{desired_formats.inspect}"
+    end
+
+    factories.map { |callable_or_class| instantiate_parser(callable_or_class) }
+  end
+
+  def self.instantiate_parser(callable_or_responding_to_new)
```
tehAnswer: `callable_or_responding_to_new` => `parser`? I guess it's a more readable name.

🙃

README.md
julik: I do it all the time primarily because we don't need to have strictly-typed references in Ruby, but I do get the idea that suggesting this in a README might be a bad idea education-wise ;-)

lib/format_parser.rb
julik: Yep.

lib/format_parser.rb
julik: How would `&` help here? `&` is a bitwise AND. And (and!) everything in `.values` here must be a Set because the only place where we assign to it is in the mirror method to register. Or am I missing something?

lib/format_parser.rb
julik: Actually here no - I'm kinda hinting at the expected type of the argument. If we say `parser` it is not clear that what we want is either the parser callable or the parser factory.

lib/format_parser.rb
tehAnswer: Oh dude, I didn't express myself correctly. What I meant something like 

`@parsers&.delete(...)`

For sure it's not the `&` operator, but yeah, this syntax can help to avoid having to create an empty-default-ish object among those lines.

lib/format_parser.rb
tehAnswer: Maybe it's just a cosmetic issue to me as this method can be halved in length, and for that dimensions, maybe the hint is not needed at all. 

😉 

```ruby
def self.instantiate_parser(parser)
  return parser if parser.responds_to?(:call)
  return parser.new if parser.is_a?(Class) && parser.method_defined?(:call)
  raise ArgumentError, '...'
end
```

tehAnswer: Also, it could mean that a code comment is needed hehehe.

julik: Or a monad 😆

lib/format_parser.rb
julik: Ah good one! I still have to get used to that trick.

PR #67, opened 2018-02-12
CR2 file support

onurkucukkece:
- Added CR2 parser
- Added preview feature for CR2 files
- Updated format parser class, refactored magical numbers to constants
- Updated README

### DIFF
--- a/lib/image.rb
+++ b/lib/image.rb
@@ -13,6 +13,9 @@ module FormatParser
     # Number of pixels vertically in the pixel buffer
     attr_accessor :height_px
 
+    # Image resolution
+    attr_accessor :resolution
+
     # Whether the file has multiple frames (relevant for image files and video)
     attr_accessor :has_multiple_frames
 
@@ -43,6 +46,9 @@ module FormatParser
     # If a parser wants to provide any extra information to the caller
     # it can be placed here
     attr_accessor :intrinsics
+
+    # Preview image blob
+    attr_accessor :preview
 
     # Only permits assignments via defined accessors
     def initialize(**attributes)
--- a/lib/image.rb
+++ b/lib/image.rb
@@ -13,9 +13,6 @@ module FormatParser
     # Number of pixels vertically in the pixel buffer
     attr_accessor :height_px
 
-    # Image resolution
-    attr_accessor :resolution
-
     # Whether the file has multiple frames (relevant for image files and video)
     attr_accessor :has_multiple_frames
 
--- a/lib/image.rb
+++ b/lib/image.rb
@@ -44,9 +44,6 @@ module FormatParser
     # it can be placed here
     attr_accessor :intrinsics
 
-    # Preview image blob
-    attr_accessor :preview
-
     # Only permits assignments via defined accessors
     def initialize(**attributes)
       attributes.map { |(k, v)| public_send("#{k}=", v) }
--- a/lib/parsers/cr2_parser.rb
+++ b/lib/parsers/cr2_parser.rb
@@ -11,7 +11,7 @@ class FormatParser::CR2Parser
   EXIF_OFFSET_TAG = 0x8769
   MAKERNOTE_OFFSET_TAG = 0x927c
   AFINFO_TAG = 0x0012
-  AF2INFO_TAG = 0x0026
+  AFINFO2_TAG = 0x0026
   CAMERA_MODEL_TAG = 0x0110
   SHOOT_DATE_TAG = 0x0132
   EXPOSURE_TAG = 0x829a
@@ -42,16 +42,16 @@ class FormatParser::CR2Parser
     set_photo_info(io, exif_offset[0])
 
     makernote_offset = parse_ifd(io, exif_offset[0], MAKERNOTE_OFFSET_TAG)
-    af_info = parse_ifd(io, makernote_offset[0], AF2INFO_TAG)
 
     # Old Canon models have CanonAFInfo tags
     # Newer models have CanonAFInfo2 tags instead
     # See https://sno.phy.queensu.ca/~phil/exiftool/TagNames/Canon.html
+    af_info = parse_ifd(io, makernote_offset[0], AFINFO2_TAG)
     unless af_info.nil?
-      parse_new_model(io, af_info[0], af_info[1])
+      parse_dimensions(io, af_info[0], af_info[1], 8, 10)
     else
       af_info = parse_ifd(io, makernote_offset[0], AFINFO_TAG)
-      parse_old_model(io, af_info[0], af_info[1])
+      parse_dimensions(io, af_info[0], af_info[1], 4, 6)
     end
 
     FormatParser::Image.new(
@@ -84,18 +84,11 @@ class FormatParser::CR2Parser
     sequence.reverse.unpack('H*').join.hex
   end
 
-  def parse_new_model(io, offset, length)
+  def parse_dimensions(io, offset, length, w_offset, h_offset)
     io.seek(offset)
     items = safe_read(io, length)
-    @width = parse_sequence_to_int items[8..9]
-    @height = parse_sequence_to_int items[10..11]
-  end
-
-  def parse_old_model(io, offset, length)
-    io.seek(offset)
-    items = safe_read(io, length)
-    @width = parse_sequence_to_int items[4..5]
-    @height = parse_sequence_to_int items[6..7]
+    @width = parse_sequence_to_int items[w_offset..w_offset + 1]
+    @height = parse_sequence_to_int items[h_offset..h_offset + 1]
   end
 
   def parse_ifd_0(io, offset)
--- a/lib/parsers/cr2_parser.rb
+++ b/lib/parsers/cr2_parser.rb
@@ -47,7 +47,7 @@ class FormatParser::CR2Parser
     # Old Canon models have CanonAFInfo tags
     # Newer models have CanonAFInfo2 tags instead
     # See https://sno.phy.queensu.ca/~phil/exiftool/TagNames/Canon.html
-    if !af_info.nil?
+    unless af_info.nil?
       parse_new_model(io, af_info[0], af_info[1])
     else
       af_info = parse_ifd(io, makernote_offset[0], AFINFO_TAG)
@@ -70,14 +70,14 @@ class FormatParser::CR2Parser
     io.seek(offset)
     entries_count = parse_sequence_to_int safe_read(io, 2)
     entries_count.times do
-      entry = safe_read(io, 12)
-      tag_id = parse_sequence_to_int entry[0..1]
-      type = parse_sequence_to_int entry[2..3]
-      length = parse_sequence_to_int entry[4..7]
-      value = parse_sequence_to_int entry[8..11]
-      return [value, length, type] if tag_id == searched_tag
+      ifd = ifd_entry safe_read(io, 12)
+      return [ifd[:value], ifd[:length], ifd[:type]].map { |b| parse_sequence_to_int b } if ifd[:tag] == [searched_tag].pack('v')
     end
     nil
+  end
+
+  def ifd_entry(binary)
+    { tag: binary[0..1], type: binary[2..3], length: binary[4..7], value: binary[8..11] }
   end
 
   def parse_sequence_to_int(sequence)
--- a/lib/parsers/cr2_parser.rb
+++ b/lib/parsers/cr2_parser.rb
@@ -6,7 +6,6 @@ class FormatParser::CR2Parser
 
   PREVIEW_ORIENTATION_TAG = 0x0112
   PREVIEW_RESOLUTION_TAG = 0x011a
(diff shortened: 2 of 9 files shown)


bermannoah:
Hey, really really nice work! 

As far as the preview sizes are concerned: is the issue that the preview images themselves do not contain the metadata, or is it that the CR2 data is missing that info? If the preview images themselves contain the relevant metadata, it could be worthwhile seeing if we can feed them back into format_parser to retrieve the info. 

Also: can you add a test to the TIFF parser spec to make sure that we don't accidentally parse our TIFF fixtures as CR2s? (I know you test that the CR2 parser doesn't do this, but it'd be nice to see what happens going the opposite direction.)

onurkucukkece:
Thank you very much. Frankly, I haven’t checked the preview image metadata. Preview sizes are missing in CR2 files itself which shot with some old Canon models. I can try to parse them from the preview image metadata however there is a byte limit for reading and seeking. Most of the CR2 files are exceeding that limit for the preview image which returns nil in that case. Shall I just parse them where available?

You are right, it would be good to have CR2 specs for TIFF parser. I will work on that as well :)

bermannoah:
Hmm. I think in that case we might be better off making the preview image metadata gathering an issue all on its own, outside the scope of this issue -- but maybe @julik has another idea? Otherwise: your current implementation - parse it from the CR2 where available - should be fine. 

And: thank _you_ very much!

onurkucukkece:
Apparently, TIFF parser was trying to parse the CR2 files :grin:, Thank you for pointing it out. I have fixed it and added its specs.

Thank you as well

bermannoah:
Ahah, glad I asked! 😁 Thanks for checking into (and fixing) it! Summoning @julik to take a look as well.

lib/parsers/cr2_parser.rb
```diff
@@ -0,0 +1,119 @@
...
+    makernote_offset = parse_ifd(io, exif_offset[0], 0x927c)
+    af_info = parse_ifd(io, makernote_offset[0], 0x0026)
+
+    # Old Canon models have CanonAFInfo tags (0x0012)
+    # Newer models have CanonAFInfo2 tags (0x0026) instead
+    # See https://sno.phy.queensu.ca/~phil/exiftool/TagNames/Canon.html
+    if !af_info.nil?
+      parse_new_model(io, af_info[0], af_info[1])
+    else
+      af_info = parse_ifd(io, makernote_offset[0], 0x0012)
```
julik: There are constants for some offsets/tag IDs, but not for others, is that intentional?

lib/parsers/cr2_parser.rb
```diff
@@ -0,0 +1,119 @@
...
+    tiff_bytes = tiff_header[0..3].bytes
+    magic_bytes = safe_read(io, 4).unpack('C4')
+
+    return if !magic_bytes.eql?(CR2_HEADER) || !tiff_bytes.eql?(TIFF_HEADER)
+
+    # Offset to IFD #0 where the preview image data is located
+    # For more information about CR2 format,
+    # see http://lclevy.free.fr/cr2/
+    # and https://github.com/lclevy/libcraw2/blob/master/docs/cr2_poster.pdf
+    if0_offset = tiff_header[4..7].reverse.bytes.collect { |c| c.to_s(16) }.join.hex
```
julik: Isn't this line doing the same as `to_hex` below?

lib/parsers/cr2_parser.rb
```diff
@@ -0,0 +1,119 @@
...
+      entry = safe_read(io, 12)
+      tag_id = to_hex(entry[0..1])
+      length = to_hex(entry[4..7])
+      value = to_hex(entry[8..11])
+      return [value, length] if tag_id == searched_tag
+    end
+    nil
+  end
+
+  def to_hex(sequence)
```
julik: Is this method really doing `to_hex` (converting to hex)? It seems that what it actually does is a "bin to hex" conversion, but it's output is an integer - and it applies an integer conversion on top. It can also be rewritten to avoid string allocations that it does right now:

```ruby
[47] pry(main)> '100'.reverse.unpack('H*').join.hex
=> 3158065
[48] pry(main)> '100'.reverse.unpack('H*').join.hex
```

Doesn't seem we can ditch the `reverse` because apparently there is no flipped byte order hex specifier for `pack`. Regardless I would give the method a better name, what could it be?

spec/format_parser_spec.rb
```diff
@@ -101,4 +101,18 @@
       }.to raise_error(/No parsers provide/)
     end
   end
+
+  describe 'maximum bytes variables' do
```
julik: To be honest I would not check that constants are being used and I would not check their constants. We can verify behavior, and if behavior is proven to be correct it probably is all we care about. The constants we extract for readability, and for testing readability we have Rubocop already policing us :-)

spec/parsers/cr2_parser_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        expect(parsed.format).to eq(:cr2)
+
+        expect(parsed.width_px).to be_kind_of(Integer)
+        expect(parsed.width_px).to be > 0
+
+        expect(parsed.height_px).to be_kind_of(Integer)
+        expect(parsed.height_px).to be > 0
+
+        expect(parsed.resolution).to be_kind_of(Integer)
+        expect(parsed.resolution).to be > 0
```
julik: In the tags we _do_ read, is there any useful information that is small and can be passed as, say, a Hash, into `intrinsics`? (also see on `Image#resolution` above)

spec/parsers/cr2_parser_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    it 'is able to return the orientation nil for the examples from old Canon models' do
+      file = fixtures_dir + '/CR2/_MG_8591.CR2'
+      parsed = subject.call(File.open(file, 'rb'))
+      expect(parsed.orientation).to be_nil
+      expect(parsed.image_orientation).to be_nil
+    end
+  end
+
+  describe 'is able to return class constant variables' do
```
julik: Do we ever use these values externally? If we aren't (if they "are not an API") why would we test them? They do affect behavior of the parser, and the parser we test as a parser.

spec/parsers/exif_parser_spec.rb
```diff
@@ -1,16 +1,21 @@
...
-  #   :top_left,
-  #   :top_right,
-  #   :bottom_right,
-  #   :bottom_left,
-  #   :left_top,
-  #   :right_top,
-  #   :right_bottom,
-  #   :left_bottom
-  # ]
+  describe 'is able to return constant variables' do
```
julik: Same. Let's not test the contents of the constants.

lib/image.rb
```diff
@@ -13,6 +13,9 @@ class Image
     # Number of pixels vertically in the pixel buffer
     attr_accessor :height_px
 
+    # Image resolution
+    attr_accessor :resolution
```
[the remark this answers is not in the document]
onurkucukkece: I realized I wasn't parsing the resolution correctly and it is fixed now. However again for some Canon models, resolution unit tag is missing in CR2, hence I'm moving it into intrinsics.

lib/image.rb
```diff
@@ -44,6 +47,9 @@ class Image
     # it can be placed here
     attr_accessor :intrinsics
 
+    # Preview image blob
+    attr_accessor :preview
```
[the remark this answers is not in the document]
onurkucukkece: Offset and byte size of preview image depends on the size and seek limits that is set in the FormatParser class. If its offset and size is bigger than the limits, preview returns nil. Would you still like me to remove it?

lib/parsers/cr2_parser.rb
onurkucukkece: This is a much better approach. Thank you very much for that. I'l try to come up with a better name for ```to_hex```method.

lib/parsers/cr2_parser.rb
```diff
@@ -0,0 +1,119 @@
+class FormatParser::CR2Parser
+  include FormatParser::IOUtils
+
+  # Variables
```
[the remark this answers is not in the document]
onurkucukkece: Indeed, it should be constants :)

lib/parsers/cr2_parser.rb
onurkucukkece: All tags moved to constants.

spec/parsers/cr2_parser_spec.rb
onurkucukkece: Camera model, shoot date, exposure and aperture values are provided in intrinsics, unfortunately I wasn't able to provide ISO number because some models are missing the ISO number tag.

onurkucukkece:
The website where I downloaded the CR2 files does not provide any license information. I will try to find other licensed examples.

onurkucukkece:
I'm sorry, I found the legal information in the site where I downloaded the CR2 examples. Updated readme accordingly.

onurkucukkece:
It sounds like good idea to gather up TIFF/CR2 and even NEF parsers in one class however I also think it would be better to make it in a different PR.

lib/image.rb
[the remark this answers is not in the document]
julik: To be very honest yes, primarily due to the intended use of the library. It is made to scan a lot of files very quickly, and we want to get information that will be useful for display but no actual pixel/audio data, since it will end up discarded anyway - we use other tools (ElasticTranscoder, vips and ImageMagick) to do the actual processing. If we return this data to the caller it will take longer to garbage collect it, it will allocate a Ruby string for each preview we pass, and all of the above won't end up being used by the calling code. So if we want to return that information in some useful form IMO we should include the byte offset at which to find the preview blob and it's size in the intrinsics, but not the actual binary string with the preview.

lib/parsers/cr2_parser.rb
```diff
@@ -0,0 +1,173 @@
...
+  PREVIEW_IMAGE_OFFSET_TAG = 0x0111
+  PREVIEW_IMAGE_BYTE_COUNT_TAG = 0x0117
+  EXIF_OFFSET_TAG = 0x8769
+  MAKERNOTE_OFFSET_TAG = 0x927c
+  AFINFO_TAG = 0x0012
+  AF2INFO_TAG = 0x0026
+  CAMERA_MODEL_TAG = 0x0110
+  SHOOT_DATE_TAG = 0x0132
+  EXPOSURE_TAG = 0x829a
+  APERTURE_TAG = 0x829d
```
julik: 👍

lib/parsers/cr2_parser.rb
```diff
@@ -0,0 +1,173 @@
...
+    @resolution = resolution_data[0] / resolution_data[1]
+  end
+
+  def set_preview(io, offset)
+    @preview_offset = parse_ifd(io, offset, PREVIEW_IMAGE_OFFSET_TAG).first
+    @preview_byte_count = parse_ifd(io, offset, PREVIEW_IMAGE_BYTE_COUNT_TAG).first
+  end
+
+  def parse_preview_image(io)
+    return if @preview_byte_count > FormatParser::MAX_BYTES || @preview_offset > FormatParser::MAX_SEEKS
```
julik: See above re. preview - in this case this will also not quite work like you expect it. The MAX_BYTES constraint is for the overall amount the parser is permitted to read from the IO in total. If we have already read some bytes for other tags we are already under that limit actually, and even though here `@preview_offset` might be lower than `MAX_SEEKS` you will probably be below `MAX_SEEKS` already for this parser. So it will in fact cause an exception to be raised if we try to read MAX_BYTES at this stage - the limited IO simply won't let us.

The reason we've set it up this way (not have parsers bother with how much they are allowed to read) – is this: if we encourage it (checking the amount of data read in parsers, in every parser), every parser then needs to have it's own arithmetic for figuring out how much it did read already, or whether it is a child of another parser, there will be bugs in that arithmetic, and we might be increasing our possible attack surface since we handle user-supplied files here. Instead we set a hard limit for every parser in one place, once, and make it globally configurable.

lib/parsers/cr2_parser.rb
onurkucukkece: I have renamed```to_hex```method to ```parse_sequence_to_int```

lib/image.rb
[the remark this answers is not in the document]
onurkucukkece: Oh, now I see what you mean. I will update the class accordingly.

lib/parsers/cr2_parser.rb
onurkucukkece: You are right, I haven't thought it that way. After moving preview data into intrinsics, ```parse_preview_image```method is not needed anymore.

julik:
@bermannoah unless you have reservations I would like to merge

julik:
Thank you @onurkucukkece !

onurkucukkece:
Thank you as well. It was quite fun. 😄

PR #69, opened 2018-02-25
Add default as_json

### LINKED ISSUE #40: A sane default FileInformation#as_json
We currently do not have any decent default conversion for `JSON.dump(file_information)` which is something we are very likely to want - for example as a result of an image upload or detection request. At the minimum we should return all the attributes that have readers and are not `nil`.

julik:
for all the supported file types, based on accessor enumeration.

Closes #40


bermannoah:
Beautiful. I bumped the version and added .as_json to the readme.

PR #71, opened 2018-02-28
Adds a basic PDF parser

westerg:
* Can detect a PDF based on the following premise:
      The first 9 bytes should contain %PDF-1.[0-8]
    * Can detect page counts if the page count is either
      in a header called /Count or /N or /Page followed
      by a number.


lib/parsers/pdf_parser.rb
```diff
@@ -0,0 +1,49 @@
...
+      page_count: page_count(io)
+    )
+  end
+
+  private
+
+  def page_count(io)
+    io.seek(9)
+
+    while read = safe_read(io, 100)
```
WJWH: - `safe_read` will raise an exception if you do not read the exact amount of bytes requested, so if the filesize is not an exact multiple of 100 (+9 to account for the PDF_MARKER), this will blow up if nothing in the document matches the COUNT_MARKER.
- Further, this does a potentially unbounded number of reads. This might be a problem if someone tries to determine the format of `malicious_20_gb_long.pdf`. We should probably have some way to limit the total amount read.
- If the COUNT_MARKER is spread across read chunks, this will not pick it up and we'll keep reading 100 byte chunks until we hit EOF.

lib/parsers/pdf_parser.rb
```diff
@@ -0,0 +1,49 @@
+class FormatParser::PDFParser
+  include FormatParser::IOUtils
+
+  class InvalidStructure < StandardError
+  end
+
+  # First 9 byes of a PDF should be in this format, according to:
```
WJWH: Typo: byes -> bytes

spec/parsers/pdf_parser_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
```
WJWH: This file could really use a spec for the unhappy path as well, since that would have caught several issues posted above.

lib/parsers/pdf_parser.rb
westerg: > safe_read will raise an exception if you do not read the exact amount of bytes requested, so if the filesize is not an exact multiple of 100 

I've seen this problem occur while fiddling around with the code actually :sweat_smile: . The alternative I do see work is to not use `safe_read` and just read the 100 bytes with `io.read(100)`. I'm guessing there are other smarter measures to take especially considering the following:

> Further, this does a potentially unbounded number of reads. This might be a problem if someone tries to determine the format of malicious_20_gb_long.pdf. We should probably have some way to limit the total amount read.

Yes, we should. What is an acceptable amount of reads? 

> If the COUNT_MARKER is spread across read chunks, this will not pick it up and we'll keep reading 100 byte chunks until we hit EOF.

I'm not sure if I fully understand this, you mean if the COUNT_MARKER is concealed within a serie of bytes? Like:

```
AAAAAAAAAAAAAAAA\Count 5
```

Instead of:

```
\Count 5
```

WJWH: If we read in chunks of 100 bytes and the first bytes of the COUNT_MARKER are in the final bytes of chunk 1 and the last bytes are the first bytes of chunk 2, then it will not be picked up by the regex even though there is a valid COUNT_MARKER in the file.

According to the PDF specs, there can be many count markers in a document as well, for example for each chapter in a document. So you'd run the risk of missing the first one and then misreporting the page count of one of the chapters as the count of the whole thing.

I think it'd be reasonable to read the first 512 or so bytes 🤔

westerg:
@WJWH I've updated the PDF parser to look a bit more like other parsers. It still doesn't solve malicious, extremely large PDF files and there is a lot more to test for with the specs, but for parsing simple PDF's I think it will do all right. Also, I believe uploading a maliciously large JPG file will be equally harmful in the current state of affairs and for the sake of consistency it might be worthwhile to figure this out on a higher level than on a single parser, but that's just my 2 cents.

WJWH:
My apologies for the delay in re-reviewing this PR. This is absolutely a good start for the PDF parser and IMO the assignment was completed successfully. Lizzie is not at the office until Tuesday sadly, but I'll ask her to contact you when she gets back 🙂

PR #72, opened 2018-02-28
Pagefault limit

### LINKED ISSUE #41: Care::IOWrapper should limit the number of pages read per parser
With the current ReadLimiter setup it is possible to have a file exploit the page cache for formats that require sequential reading of chunks to parse out metadata (TIFF, JPEG, MOVs etc. in the future). The exploit goes like this: a file has to be crafted which has "fake chunks" stating sizes just the size of the IOWrapper page or slightly less / slightly more. 

For example, we can create a MOV file in #36 which is going to be filled with `\x00\x00@\x00YOLO` followed by 64 kilobytes of garbage. This will force the parser to load the start section of each "YOLO atom" and skip over to the next - since it doesn't know how to deal with the content. That, in turn, is going to load the next page - also containing `\x00\x00@\x00YOLO` followed by 64 kilobytes of garbage, and so ad infinitum. 

Basically, an exploit using this will direct the parser to read chunks in succession, with each chunk being located just within the subsequent page, which in turn is going to incur a page load into the cache (page fault). Given enough chunks, a parser can be fooled into requiring 1 byte from each page and thus force the entire file's contents to be loaded into memory, resulting in a subtle DoS - as well as incurring many extra HTTP requests in case of remote resource parsing.

Mitigation strategies (I'd say in combination rather than separately)

* Jittering the cache page size for each `parse()` call, so that the cache page size is less predictable. This makes the attack less convenient to execute since you don't know how to size your chunks in advance. The Care cache store supports this already with arbitrary page sizes
* Jittering page size per page. We could do that but this will complicate arithmetic within Care quite a bit
* Limiting how many page faults a parser might incur per run. For example, we can say that "any parser is not permitted to incur more than 3 page faults", which, being an arbitrary number, should at least cap the number of extra HTTP requests a parser will be able to incur, which in the worst case scenario will be 3*number of parsers that parse in chunks. We can do this by sliding another ReadLimiter just under the Care::IOWrapper (yes, it is turtles all the way down).

julik:
Forbid parsers to cause more than N page faults, which when using RemoteIO will turn into more and more HTTP requests.

Also remove the initial read() for remote data since we would have to perform it anyway

Closes #41

### DIFF
--- a/format_parser.gemspec
+++ b/format_parser.gemspec
@@ -39,5 +39,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'simplecov', '~> 0.15'
   spec.add_development_dependency 'pry', '~> 0.11'
   spec.add_development_dependency 'yard', '~> 0.9'
-  spec.add_development_dependency 'wetransfer_style', '0.4.0'
+  spec.add_development_dependency 'wetransfer_style', '0.5.0'
 end
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -13,9 +13,11 @@ module FormatParser
   require_relative 'care'
 
   PARSER_MUX = Mutex.new
-  MAX_BYTES = 512 * 1024
-  MAX_READS = 64 * 1024
-  MAX_SEEKS = 64 * 1024
+
+  MAX_BYTES_READ_PER_PARSER = 512 * 1024
+  MAX_READS_PER_PARSER = 64 * 1024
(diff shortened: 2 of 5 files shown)


linkyndy:
@julik there was an [issue](https://github.com/bbatsov/rubocop/issues/5224) with the version of `rubocop` that we were using here. Luckily, we already have the newest version supported by `wetransfer_style`, so upgrading to it solves the problem.

PR #77, opened 2018-03-18
Make sure root: and to_json without arguments work

julik:
It is possible to call to_json without passing
a generator object, which only gets passed if
the to_json is called by JSON.dump/JSON.pretty_generate

Also implements sane support for the root option as
per the ActiveModel::Serializers convention


julik:
@roysab ?

PR #80, opened 2018-03-24
Remove parse_atom_children_and_data_fields

### LINKED ISSUE #63: parse_atom_children_and_data_fields not implemented in the MOOV Decoder
It is a method meant to deal with atoms that contain both data fields _and_ child atoms, which are few and far between - but do exist of course. I left a stub for handling these since this is where the useful metadata lives in .M4A but didn't write the implementation, the result being that the parser breaks.

Immediate mitigation is basically removing this case in the parser and pretending these atoms are opaque to us (skipping over them).

bermannoah:
I feel no pride in doing this 😓but for the sake of squashing a persistent and unnecessary error elsewhere, let's (temporarily) disable parse_atom_children_and_data_fields and create a new issue/alter the existing issue for parsing atoms that contain both data fields and child atoms.

Potentially closes #63.


julik:
There is no test case for this, and that AAC file you promised to get from iTunes would make a perfect one 🤔 since it has those AAC mixed-type boxes for the metadata

bermannoah:
Whoops. I added a fixture but when I add it to the tests it does kind of half-parse it and return a MOV. 🤦‍♂️ Feel like it's probably just best to figure out the parse_atom_children_and_data_fields method. I'll take a look either later this weekend or next week.

PR #83, opened 2018-03-27
Add parse_atom_children_and_data_fields support 

### LINKED ISSUE #63: parse_atom_children_and_data_fields not implemented in the MOOV Decoder

bermannoah:
Will close #63.

A bit of a hack to convince the decoder to keep following the atom children / tree until it has the data it needs. Also, since M4A only supports audio we need to change the nature + data returned from this filetype only.


spec/parsers/moov_parser_spec.rb
```diff
@@ -34,17 +34,24 @@ def deep_print_atoms(atoms, output, swimlanes = [])
       result = subject.call(File.open(moov_path, 'rb'))
 
       expect(result).not_to be_nil
-      expect(result.nature).to eq(:video)
-      expect(result.width_px).to be > 0
-      expect(result.height_px).to be > 0
+      expect(result.nature).to eq(:video) unless result.format == :m4a
```
julik: If the test is generic ("for al MOOVs")  we should also have generic assertions. For example we can verify that `#nature` is "one of :video or :audio". But I would strongly discourage adding conditionals in tests - if we need those, we should have separate tests for specific movie files and separate tests for specific audio files.

bermannoah: Ahh, makes sense. How would you suggest handling it for the width_px and height_px fields?

julik: As soon as you have tests for each mov file, remove them in this spec :-)

lib/parsers/zip_parser/file_reader.rb
```diff
@@ -361,11 +360,13 @@ def locate_eocd_signature(in_str)
     minimum_record_size = 22
     end_location = minimum_record_size * -1
     loop do
+      # -100 here is an arbitrary number, but if not in place the this will continue
+      # reading gibberish (e.g. in our fuzz tests) forever
```
bermannoah: This will blow up our fuzz tests if it doesn't have some way to fail past a certain end location. My solution (returning if the end location is less than -100) is a hacky lil patch - @julik, is there a way we can detect if something is a ZIP before hitting this loop? I've also tried setting this to be a ReadLimiter but it doesn't appear to make a difference, since ofc it's already done the reading by this point.

julik: @bermannoah I will make the loop bounded. Also, completely unrelated at all to this PR ;-)

bermannoah: :-) No worries! Just wanted to make sure it wasn't the recursive decoder causing the issue.

bermannoah: @julik would you like me to remove the -100 here since the fix is outside the scope of the PR? It does seem to cause issues with running the tests.

julik:
Then let's open a separate PR to address it

Issue #82, opened 2018-03-27
.docx files are sometimes read as MP3s


bermannoah:
We've seen .docx files being parsed and returning `:previewable=>true, :preview_metadata=>{:duration_in_seconds=>"728.97306"}` which is curious and undesirable. Sample file in comments.

bermannoah:
[Here](https://we.tl/3R6yNOxJu8) is the sample file that caused this issue, used here with the permission of the owner.

julik:
I've added a check for the ZIP local entry signature at the start of the file to try and avoid this - the MP3 parser has to be pretty liberal due to the way MP3 files are structured.

Issue #81, opened 2018-03-27
DOC/DOCX format support


bermannoah:
Right now we just need to be able to identify them and return page count if possible. Microsoft has some [docs](https://msdn.microsoft.com/en-us/library/office/dd950942(v=office.14).aspx) to (maybe) get you started.

julik:
This needs a ZIP parser.

bermannoah:
@julik can you make an issue for what we need from a ZIP parser?

julik:
We can work on this. There is dangerous territory though - the items in those zips may - and probably will, in 100%, be _deflated_. And that, in turn, means ZIP bombs. http://www.aerasec.de/security/advisories/decompression-bomb-vulnerability.html

To get at things such as page count we need to parse the actual XML of the documents, which then means that we need a real XML parser. Which also has the billion laughs exploit.

@bermannoah can you investigate in detail where we at with deflate bombs in the Ruby zlib module please?

bermannoah:
Sure -- will investigate next week. Found some fun [sample files](https://bomb.codes/bombs) to test with -- neat image implementations in there too.

PR #86, opened 2018-03-30
Add basic detection of Office files

julik:
Adds basic MS Office file format detection on top of the ZIP parser.

Should address #81 #82


julik:
This needs example files that can be freely used with the library. @roysab @bermannoah care to throw some my way? No tests yet for the same reason.

bermannoah:
[Here](https://nb.wetransfer.com/downloads/a18751f261314fb02e2dd31c0cd9d4ef20180330090735/2fc372) is the original file that alerted us to the issue, and [here](https://we.tl/7OhUeYqrKI) is one that does not _appear_ to cause the issue -- both of which we can use with @Daniela-Vandor's permission.

julik:
@bermannoah that should do for now

lib/parsers/zip_parser/office_formats.rb
```diff
@@ -0,0 +1,51 @@
...
+# Additionally, per file type, they contain the following:
+#
+#   word/document.xml
+#   xl/workbook.xml
+#   ppt/presentation.xml
+#
+# These are sufficient to say with certainty that a ZIP is in fact an Office document.
+# Also that unscientific sample revealed that I came to dislike MS Office so much as to
+# only have 63 documents on my entire workstation.
+#
```
bermannoah: 😂

spec/esoteric_formats_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe 'Parsing esoteric files and files causing ambiguous detection' do
```
bermannoah: Is there any particular reason to do this test here and not add the spec to MP3 parser spec as well? It's not a blocker but I'm curious. (I do like having a specific place for this sort of thing!)

julik: Because I believe this is not the last time we discover a file like this. As a matter of fact we can run a contest about crafing a file that would detect as _as many_ file types as possible and all in err.

PR #96, opened 2018-04-13
Add BMP parser

### LINKED ISSUE #56: BMP parser
recovering resolution and color mode

GustavoCaso:
Fixes #56 

@julik 

I have added a BMP parser. It is my first time actually having to look at file format specification and working with them. I actually had a great time. I was finding a bit difficult to find information about the `color_mode` of the BMP.

In my implementation, I try to follow as a much a possible the other parsers.

Please add as many comments as you need I would love to address then back in the meantime I would look for more information on the subject and try to work on another parser


spec/parsers/bmp_parser_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
+
+describe FormatParser::BMPParser do
+  describe 'is able to parse all the examples of BMP' do
+    Dir.glob(fixtures_dir + '/*.bmp').each do |bmp_path|
```
julik: Let's move those into the BMP/ directory in fixtures/ - right now it would be too easy to remove the .bmp files from fixtures/ and we wouldn't notice since they are stored there together with all the other examples we imported from FastImage.

lib/parsers/bmp_parser.rb
```diff
@@ -0,0 +1,41 @@
+class FormatParser::BMPParser
+  include FormatParser::IOUtils
+
+  VALID_BMP = 'BM'
+  PIXEL_ARRAY_OFFSET = 54
+  COLOR_TYPES = {
+    1 => :monochrome
+  }
+
+  # Based on https://en.wikipedia.org/wiki/BMP_file_format
```
julik: The comment is kinda hanging in between - it doesn't belong to the class, neither does it belong to the method underneath due to the spacing. Maybe we could put it at the top of the file?

lib/parsers/bmp_parser.rb
```diff
@@ -0,0 +1,41 @@
+class FormatParser::BMPParser
+  include FormatParser::IOUtils
+
+  VALID_BMP = 'BM'
+  PIXEL_ARRAY_OFFSET = 54
+  COLOR_TYPES = {
+    1 => :monochrome
```
julik: Could the data available here be useful? http://www.dragonwins.com/domains/getteched/bmp/bmpfileformat.htm#The%20Color%20Table

GustavoCaso: @julik after some read and some experimentation by using ImageMagick to convert `png` with alpha to `bmp` and checking the content of the `io` object and been the same I decide to make the color mode `:rgb` by default. It would be much faster since there is no need to read more from the `io` object. Please correct me if I'm wrong. Thanks

spec/parsers/bmp_parser_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
+
+describe FormatParser::BMPParser do
+  describe 'is able to parse all the examples of BMP' do
+    Dir.glob(fixtures_dir + '/BMP/*.bmp').each do |bmp_path|
```
julik: Almost. Given that we only have 2 specific BMP fixtures, would it maybe make sense to parse them in separate examples and check that the pixel dimensions are correct? We are verifying that they are Integers, which is fine and dandy but they might be `-5` if the parser fails somewhere. We can't verify it in this test since it is automatically created by globbing the fixtures dir.

GustavoCaso: Excellent point. Actually, I found that the height could be a negative number that indicates that the data order is from `ordered from top to bottom` the link is this one http://www.dragonwins.com/domains/getteched/bmp/bmpfileformat.htm#The%20Image%20Header in the section `biHeight `. In the case of a negative number we could actually return the `abs` of it and in the `intrinsics` section that the data is ordered inverse

julik: Wow. I only speculated this could happen to prove a point but I could never in my wildest dreams imagine this could **actually** be the case. Yes, I believe it's better to give positive image dimensions to callers regardless of the BMP scan order 😆 amazing gem of a find

GustavoCaso: Actually, the `test2.bmp` file has negative height 😆 so it was serving our case perfectly

julik:
👍 🚀 🎉

PR #98, opened 2018-04-14
Fix EXIF parsing in JPEG files

julik:
When parsing JPEGs via EXIFR what EXIFR would do is actually the following: it would seek to the start of the JPEG, scan each marker (just like we do), scan the SOF markers (just like we do), then find the APP1 marker (just like we do) and then try to read that and parse the contents of the APP1 marker as if it came from a TIFF file. Then it would scan _forwards_ until the EOI marker, thereby actually skipping over the entire file - and possibly slurping it in to jump to the next marker.

This, in turn, would cause us to perform at least twice as many read() calls on our IO as necessary, which triggers the read limiter in some edge cases since there would be so many.

We also can't use the opportunity to read all the EXIF data from upstream using a single read() - which, APP1 being a single unit of data within the file - should be perfectly possible.

We remove all the JPEG-specific stuff from our EXIFR adapter and pass it the "TIFF-ish" contents of the APP1 marker to deal with, which should remove that problem.

The fact that we encountered so many problems with JPEGs causing too many reads can be directly attributed to this misuse of EXIFR.

Keynote files turn out to be a peculiar case - they start as if they were JPEG files (SOI marker, APP1 marker and all) and end with the ZIP contents of the actual Keynote bundle. To at least detect that we bail out of the JPEG parser if we are looking at a ZIP, and we thereby let Keynote documents to fall through to our ZIP parser which tags them as archives - which, for the time being, is reasonable. After all, they are "more" of a ZIP file kind than of the JPEG file kind.

File formats.


lib/parsers/exif_parser.rb
```diff
@@ -48,22 +48,16 @@ def seek(n, seek_mode = IO::SEEK_SET)
...
     @exif_data = nil
     @orientation = nil
     @height = nil
     @width = nil
   end
 
-  def scan_image_exif
-    # Without the magic bytes EXIFR throws an error
-    @file_io.seek(0)
-    raw_exif_data = EXIFR::JPEG.new(@file_io) if @filetype == :jpeg
```
duplamatyi: if `EXIFR::JPEG` is not used anymore, than probably we can safely remove `require 'exifr/jpeg'` from the beginning of the file

lib/parsers/tiff_parser.rb
```diff
@@ -13,8 +13,8 @@ def call(io)
     return if !endianness || cr2_check(io)
 
     w, h = read_tiff_by_endianness(io, endianness)
-    scanner = FormatParser::EXIFParser.new(:tiff, io)
-    scanner.scan_image_exif
+    scanner = FormatParser::EXIFParser.new(io)
+    scanner.scan_image_tiff
```
duplamatyi: could we add a line of comment here that this method does not catch the `EXIFR::MalformedTIFF` exception on purpose?

julik: Yep

lib/parsers/exif_parser.rb
julik: Good point. Less requires = faster loading time

lib/parsers/tiff_parser.rb
julik: Actually on second thought - this was not in the code previously, and we don't have any example files with invalid EXIF in a TIFF, so I'll leave it as is for now.

duplamatyi: 👍 either way is good for me. just calling a method for throwing an exception might be a bit cryptic without comments for future maintainers.

PR #100, opened 2018-04-17
Add instrumentation

julik:
For ATC - but also for image_vise - it would be useful to see how many reads we perform, how many pagefaults, and how many HTTP requests we do, etcetera. We also should think about a generic way of adding multiplexed (sent to multiple destinations) instrumentation.

This PR adds an instrumentation device called `Measurometer` (we can extract it into a gem later if we desire) which can be hooked into Appsignal like so:

    FormatParser::Measurometer.drivers << Appsignal

Once you do, there will be delicious measurements and call graphs in Appsignal for everything format_parser related, as well as some counters and samples to look at. For example, how many files end up unrecognized? What is the distribution of file formats that we are getting? etc.

### DIFF
--- a/lib/care.rb
+++ b/lib/care.rb
@@ -173,9 +173,10 @@ class Care
     # @param io[IO] the IO to read from
     # @param page_i[Integer] which page (zero-based) to read
     def read_page(io, page_i)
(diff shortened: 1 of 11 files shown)


julik:
@roysab 🎉

PR #102, opened 2018-04-17
Simplify BMP parser test

julik:
Reduce the number of nested contexts. Contexts are very useful
when there are multiple examples within them and when you need
common setup/teardown operations done for each example.

In this case, however, they create more indentation in
test formatters and also in code.

Also changes dimension assertions to checking actual values,
since we want to ensure the actual dimensions we got are correct.
For test2.bmp I could only assume the values are correct
since no application I got actually opens it. Should have used
GraphicConverter :-P

### DIFF
--- a/spec/parsers/bmp_parser_spec.rb
+++ b/spec/parsers/bmp_parser_spec.rb
@@ -1,49 +1,39 @@
 require 'spec_helper'
 
 describe FormatParser::BMPParser do
-  describe 'is able to parse BMP files' do
-    context 'with positive height_px values' do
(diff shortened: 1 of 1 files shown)


julik:
@GustavoCaso I simplified the test a little, can you give a quick 👍 👎 ? You are not on the maintainers list so not assigning review.

GustavoCaso:
Hi @julik It looks cleaner 👍. Also regarding the values from the `test2.bmp` when I open it with my editor I can not see a thing but the dimensions that appear in the bottom right corner match the ones in the test.

image

PR #105, opened 2018-04-19
Speed up JPEG parser for invalid inputs

julik:
Previously we could keep scanning JPEG files until reaching the read limits,
which was also slow because of the incurred read/unpack/first calls for each
byte. This could lead to parse calls for, say, PSD files take 3-4 seconds since
we would be looking for a marker and not finding it and would just keep reading
one byte at a time.

This will make read_next_marker fail earlier if it cannot find a marker sequence
after "blind scanning" 1024 bytes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Make JPEG parser bail out earlier if no marker can be found for 1024KB of raw data
+
 ## 0.9.0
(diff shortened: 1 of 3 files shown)


julik:
See the profile. This is a PSD file and it takes an inordinate amount of time for us to "drop through" to the PSD parser, which captures the file metadata quickly.

perf_profile

PR #106, opened 2018-04-20
Fix parsing of JPEGs with lots of APP1 markers, fix JPEG/MP3 mislabeling

julik:
We would read each APP1 marker as if it were useful and then determine whether this APP1 marker was EXIF. What we should do instead is read the first N bytes of the marker, check if it is EXIF, and only then proceed to read the entirety of the marker for passing to EXIFR. This spares on reads.

We also have a problem where some files get mislabeled as MP3 and the metadata gets returned with a duration of Float::INFINITY. When this happens we should instead avoid returning a result altogether, since this is most likely not an MP3 at all.

And we need to bump the number of permitted pagefaults again so that we can churn through those JPEGs with lots of APP1, since reading them unfortunately happens page-per-page even if we read from those pages economically

### DIFF
--- a/lib/attributes_json.rb
+++ b/lib/attributes_json.rb
@@ -20,6 +20,7 @@ module FormatParser::AttributesJSON
     methods.grep(/\w\=$/).each_with_object(h) do |attr_writer_method_name, h|
       reader_method_name = attr_writer_method_name.to_s.gsub(/\=$/, '')
       value = public_send(reader_method_name)
+      value = nil if value == Float::INFINITY
       # When calling as_json on our members there is no need to pass the root: option given to us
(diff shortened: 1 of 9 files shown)


lib/attributes_json.rb
```diff
@@ -20,6 +20,7 @@ def as_json(root: false)
     methods.grep(/\w\=$/).each_with_object(h) do |attr_writer_method_name, h|
       reader_method_name = attr_writer_method_name.to_s.gsub(/\=$/, '')
       value = public_send(reader_method_name)
+      value = nil if value == Float::INFINITY
```
bermannoah: 🙌

PR #107, opened 2018-04-21
Add a binary called `format_parser_inspect`

julik:
which allows you to get output from format_parser without doing it from code. This is primarily useful when debugging format detection issues, but can also be used as a shortcut to determine what the file actually is. Think about it as $file on steroids.

It will print detection information on STDOUT in JSON, so `jq` can be used on the output. Use it like so:

```
$ ./exe/format_parser_inspect /path/to/somefile.jpg
```


julik:
@roysab

PR #110, opened 2018-04-22
S - Add "display" dimensions in addition to pixel buffer dimensions

julik:
In many situations you want to know "What is the resolution of this image with all the EXIF rotations and non-square pixel ratios applied" as opposed to "What is the resolution of the pixel buffer of this image".

IMO we should consider this eventuality and provide conveniences for formats which allow divergent pixel aspects / orientation tagging.


julik:
@bermannoah We could, but at this stage the modules used for it are not even documented (and not tested outside of the DPX parser itself). They also cover a pretty specific narrow use case which most of the formats do not have - namely parsing a header which is _always_ fixed size. Most formats are actually structured with chunks of variable size, with the chunk size being encoded after the chunk identifier (PNG, JPEG, MOOVs..., AIFFs...)

PR #111, opened 2018-05-01
The curious case of multiple EXIFs per image

julik:
According to https://github.com/sdsykes/fastimage/issues/102 the first EXIF in the image should be used for orientation. We, however, do not do that and yet we somehow end up with the correct orientation? Researching...


julik:
@renchap do you mind us using your test file for experimentation?

julik:
What matters is that we detect the pixel dimensions right, so will close and keep the file (and a test). In practice we use the _last_ EXIF APP1 segment in it's entirety, and _only_ that last segment.

PR #112, opened 2018-05-02
Make tests pass with JRuby

### LINKED ISSUE #99: jRuby seems to have broken `undefined: replace` behavior with String#encode
which makes it fail multiple of our tests:

https://travis-ci.org/WeTransfer/format_parser/jobs/367113788

While we do not use jRuby ourselves I think it's prudent to fix this

kostyadubinin:
Closes #99

- [x] Fix a typo in the zip parser that causes tests to fail with JRuby.
- [x] Add support for Ruby 2.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,9 @@ rvm:
 - 2.3.0
 - 2.4.2
 - 2.5.0
-- jruby-9.0
+- jruby-9.1
 sudo: false
 cache: bundler
-matrix:
-  allow_failures:
-    - rvm: jruby-9.0
 script:
   - bundle exec rubocop
   - bundle exec rspec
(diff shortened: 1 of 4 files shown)


julik:
Good stuff 👌 What if we remove the safe navigation operator and replace it with something more old-fashioned - will we then become compatible with 9.0? We can then reintroduce 2.2 compatibility as well?

kostyadubinin:
Hey Julik, I will check if the safe navigation is the only issue.

kostyadubinin:
Safe navigation turned out to be the only issue with Ruby 2.2 and JRuby 9.0. Fixed in 02d84808260eedfc4abbcc0412ecad9c483d1887

Also I added the `Unreleased` section to the changelog, but if you don't use it I can remove it.

julik:
👍 cool. Thanks for the PR!

Issue #91, opened 2018-04-09
All strings coming from the parsers (in non-intrinsics at least) must be valid UTF-8


julik:
to prevent nasty display shenanigans and problems with JSON-encoding them

PR #114, opened 2018-05-07
Make sure JSON output strings are sanitized

julik:
Make sure we only output UTF-8 data in JSON generated from the FormatParser
results.

The reason we need to do it in this (rather elaborate) way is this: `JSON.generate` and
friends use a totally different code path than `"foo".to_json(generator_state)`. We cannot predict
which one of these two ways our users will be using, and at the same time we need to prevent
invalid Strings (ones which cannot be encoded into UTF-8) as well as Float::INFINITY values
from being passed to the JSON encoder. Since we cannot override the JSON generator with
these additions, instead we will deep-convert the entire object being output to make sure
it is up to snuff.

All of the output of AttributesJSON will be sanitized this way.


julik:
Closes #91

Issue #45, opened 2018-01-09
OGG parser


julik:

kostyadubinin:
@linkyndy I will take a look into this one. Want to try to add support for Ogg Vorbis. Would it be a good start?

linkyndy:
@kostyadubinin sure, good luck with it!

PR #116, opened 2018-05-08
Add support for the Ogg format

kostyadubinin:
Fixes WeTransfer/format_parser#45

Tested only with Ogg Vorbis.
The sample is generated with AWS Polly.

The parser fetches:

- audio_sample_rate_hz
- media_duration_seconds
- num_audio_channels

In order to calculate the duration the file has to be sought page by page till the end.

Hey @linkyndy, please take a look.


julik:
It's up to @linkyndy to give input on that, but I suspect there might be something similar to the xing header in Ogg. Additionally, it will probably blow through the page cache limiter (for very good reasons!) if it has to read, say, a thousand frames from a long audio file. Maybe something like what we do for MP3 files is in order here (thinking out loud)

kostyadubinin:
It doesn't look like Ogg has anything similar to the xing header, but I found out that the size of a usual Ogg page is 4kB-8kB and maximum is 65307 bytes.

Is reading 8kB from the end would be still too much?

Comparing to the MP3 parser it would be much less. Parsing `spec/fixtures/MP3/Cassy.mp3` requires reading 179139 bytes of data and performing 133 reads and 134 seeks. The Ogg parses would require reading only 65307 bytes in the worst case and only few reads and seeks.

julik:
> Is reading 8kB from the end would be still too much?

Reading from the end of the file is fine - the IO object that you get has `size` and `seek` for that, reading from the end of the page is fine too - the fine line here is, I guess, _stopping_ with reading after  N pages and making an estimate of how long the file is given the size of the pages seen so far (and how many samples they contain, their sample rate etc...)

julik:
> The Ogg parses would require reading only 65307 bytes in the worst case and only few reads and seeks.

That should work fine and not bust the limit. I would recommend adding a _loooooooong_ ogg audio file to verify ;-)

julik:
@kostyadubinin where we at?

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+    #   2) the characters v','o','r','b','i','s' as six octets
+    packet_type, vorbis, _vorbis_version, channels, sample_rate = io.read(16).unpack('Ca6LCL')
+    return unless packet_type == 1 && vorbis == 'vorbis'
+
+    # Read the last page of the audio in order to calculate the duration.
+    pos = io.size - MAX_POSSIBLE_PAGE_SIZE
+    pos = 0 if pos < 0
+    io.seek(pos)
+    page = io.read(MAX_POSSIBLE_PAGE_SIZE)
+    pos_of_the_last_page = page.rindex('OggS')
```
julik: Is it guaranteed that these magic bytes do not occur within the page itself? If it is, you might encounter edge cases. The ZIP reader has a similar usage pattern (find a sequence of magic bytes + some data afterwards at the end of the file) and I had to implement this https://github.com/WeTransfer/format_parser/blob/master/lib/parsers/zip_parser/file_reader.rb#L347 to do it reliably.

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+    header = page[pos_of_the_last_page..pos_of_the_last_page + 13]
```
julik: If for whatever reason the file ends with garbage, `rindex` will return `nil` and you will be doing a nil-address on the string, which will probably raise a NoMethodError or similar. You probably need a bailout return where `rindex` might return nil - also see the previous comment.

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    _capture_pattern, _version, _header_type, granule_position = header.unpack('a4CCQ')
```
julik: Is 13 the size of `a4CCQ`?

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+    # begins with the characters, "OggS", to identify the file as Ogg format.
+    capture_pattern = io.read(4)
+    return unless capture_pattern == 'OggS'
+
+    io.seek(28) # skip not important bytes
+
+    # Each header packet begins with the same header fields.
+    #   1) packet_type: 8 bit value (the identification header is type 1)
+    #   2) the characters v','o','r','b','i','s' as six octets
+    packet_type, vorbis, _vorbis_version, channels, sample_rate = io.read(16).unpack('Ca6LCL')
```
julik: `L` is "32-bit unsigned, native endian (uint32_t)". That means that if this code is run on a differently-endian platform it will give you results you do not expect. Granted, few people use PowerPC these days, and ARM has switchable endianness - but do you want to leave the user at the mercy of whatever their byte order is?

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+    pos = io.size - MAX_POSSIBLE_PAGE_SIZE
+    pos = 0 if pos < 0
+    io.seek(pos)
+    page = io.read(MAX_POSSIBLE_PAGE_SIZE)
+    pos_of_the_last_page = page.rindex('OggS')
+    header = page[pos_of_the_last_page..pos_of_the_last_page + 13]
+
+    _capture_pattern, _version, _header_type, granule_position = header.unpack('a4CCQ')
+
+    duration = granule_position / sample_rate.to_f
```
julik: I would recommend doing a sanity check that `duration` is not Float::INFINITY or not epsilon-close to that - if it is it would mean something is wrong with the parsing process or the file is invalid.

lib/parsers/ogg_parser.rb
julik: And similar - `Q` is native-endian byte order. Absolute most formats specify endianness which can be hinted with the wider/newer `unpack` specifiers using `<` and `>`.

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
+# https://xiph.org/vorbis/doc/Vorbis_I_spec.pdf
+# https://en.wikipedia.org/wiki/Ogg#Page_structure
+class FormatParser::OggParser
+  # The maximum size of an Ogg page is 65,307 bytes.
+  MAX_POSSIBLE_PAGE_SIZE = 65307
```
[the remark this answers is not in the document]
julik: That can be done later. At the moment we are quite liberal with cache page sizes since performing an HTTP request with latencies and resolves and all takes longer than simply slurping in more data. The page size is generous and you will be touching the last page anyway, it will be in memory - so no worries 👍

julik:
Another tiny remark - when you do `io.read(n)` you may receive less bytes than you requested, and then your `unpack` will fail with "too few elements". To safeguard against that use `safe_read` as done in the other parsers - they will raise an exception if you receive less bytes from the IO than desired, and that exception is considered normal in the parser lifecycle - the parser just gets skipped then.

lib/parsers/ogg_parser.rb
kostyadubinin: Good question, didn't think about this. Will check.

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    # Each header packet begins with the same header fields.
+    #   1) packet_type: 8 bit value (the identification header is type 1)
+    #   2) the characters v','o','r','b','i','s' as six octets
+    packet_type, vorbis, _vorbis_version, channels, sample_rate = io.read(16).unpack('Ca6LCL')
+    return unless packet_type == 1 && vorbis == 'vorbis'
+
+    # Read the last page of the audio in order to calculate the duration.
+    pos = io.size - MAX_POSSIBLE_PAGE_SIZE
+    pos = 0 if pos < 0
```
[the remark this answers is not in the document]
kostyadubinin: Positing 0 is possible in the following two cases:

1. the file contains only one page
2. the file contains multiple small pages and the file size is less than 65307 bytes.

If the file contains only one page the last page will also be the first page which is fine.

lib/parsers/ogg_parser.rb
kostyadubinin: Ah good catch, thanks!

lib/parsers/ogg_parser.rb
kostyadubinin: > Is 13 the size of a4CCQ?

Is it wrong? I calculate it like this:
`Q` is 8 bytes
`C` is 1 byte
`a4` is 4 bytes

lib/parsers/ogg_parser.rb
kostyadubinin: Sadly it can occur, but each page contains a checksum. Will use it for verification.

lib/parsers/ogg_parser.rb
kostyadubinin: I checked the docs and I found out that Ogg doesn't specify endianness:

> The Vorbis bitpacking convention specifies storage and bitstream manipulation at the byte, not word, level, thus host word ordering is of a concern only during optimization when writing high performance code that operates on a word of storage at a time rather than by byte. Logically, bytes are always coded and decoded in order from byte zero through byte n.
https://xiph.org/vorbis/doc/Vorbis_I_spec.pdf

 I am not 100% what it means and which endianness I should use.

Is there an easy way to test the parser on a differently-endian platform?

kostyadubinin:
Hey @julik I added checksum verification. However my current implementation doesn't iterate over all occurrences of the "OggS" magic bits in the tail of the file (last 65307 bytes) yet. Please take a look at this todo comment: https://github.com/WeTransfer/format_parser/pull/116/files#diff-d43d61d8556bc303269b49b22d667dadR80

At this moment there are three things are left to do:

- [ ] finish with the checksum verification
- [ ] handle endianness
- [ ] add Float::Infinity check (https://github.com/WeTransfer/format_parser/pull/116/files#diff-d43d61d8556bc303269b49b22d667dadR85)

I will not be able to work on the parser till the end of the week. Will be back on Monday.

lib/parsers/ogg_parser.rb
julik: Nope, nothing is "wrong" I was just double-checking since the `read` and the `unpack` are not very close to each other. This should be fine as long as you get enough bytes from your read()

lib/parsers/ogg_parser.rb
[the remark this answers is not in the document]
julik: 👍

lib/parsers/ogg_parser.rb
julik: In the section you are referring to the spec apparently describes the _bitstream_ of encoded audio, and for that section their statement makes sense as you do not need to specify endianness if you read byte-by-byte.

There is however a different section that _does_ specify endianness for encoded uints, let me know if you can't find it ;-) To make an experiment without having a platform of different endianness (or to be lazy and not having to read the spec) you can check which one is "native" for the platform you wrote the tests on in `irb` even:

```
'ohai'.unpack('L') == 'ohai'.unpack('V') # If true, your platform is V-native, if not - N-native
```

julik:
@kostyadubinin That's ok, take your time.

It looks like either the Ogg CRC32 uses a different polynomial or it is encoded differently. I see that mutagen does manage to use the zlib CRC32 - the same we have in Ruby - for this by using some bit twiddling hacks, which is interesting: https://github.com/quodlibet/mutagen/blob/master/mutagen/ogg.py#L157

It can probably be ported to Ruby, but I wouldn't fret this too much yet - the setup you have with a handmade CRC is fine.

kostyadubinin:
Hi @julik I added a recursion for searching the last Ogg page bda9a7522619dd7ec335beac83e862fdc8895948, please take a look.

julik:
@kostyadubinin thanks! please address all the comments before we do the next review ;-)

lib/parsers/ogg_parser.rb
kostyadubinin: I added the sanity check like you did in the MP3 parser.

lib/parsers/ogg_parser.rb
```diff
@@ -0,0 +1,190 @@
...
+    return find_granule_position(tail[0...pos_of_the_last_page]) if header.size < 27
+
+    _capture_pattern,
+    _version,
+    _header_type,
+    granule_position,
+    _bitstream_serial_number,
+    _page_sequence_number,
+    checksum,
+    page_segments = header.unpack('a4CCQVVVC')
```
kostyadubinin: _I am copying [my previous reply](https://github.com/WeTransfer/format_parser/pull/116#discussion_r191242229) because it was lost in outdated comments._

@julik this is the last part if I didn't miss anything, but I'm stuck on it. I think it should be fine to use the native endianness, because the docs don't specify it, but I don't know how to test it on a big-endian system so not 100% sure.

The is a place in the docs that specifies endianness:

> However, since data in the comment header is octetaligned,
they can simply be read as unaligned 32 bit little endian unsigned integers.

But it is about the Vorbis text comment header which I don't use.

If you could help me with this part it would be great!

julik:
@kostyadubinin This PR was open for a _really_ long time. Do you mind if I merge it onto a different branch than master and then polish it a bit?

PR #122, opened 2018-06-08
Improve performance of the PDF parser

grdw:
**Performance of the PDF parser:**

```
#### 0. SNAFU (master branch):
              user     system      total        real
         11.549356   5.427628  16.976984 ( 17.000083)

#### 1. Improved (reading 1024 bytes):

              user     system      total        real
          0.689317   0.293554   0.982871 (  0.984320)

#### 2. Improved (reading 50 bytes instead of 1024):

              user     system      total        real
          1.107154   0.448861   1.556015 (  1.557351)

#### 3. Improved (reading 1024 bytes) + break when all MARKERS are found:

              user     system      total        real
          0.620440   0.238385   0.858825 (  0.859718)

#### 4. Improved (reading 1024 bytes) + break when all MARKERS are found + use 'index' instead of a Regex to look for a '/':

              user     system      total        real
          0.611088   0.246337   0.857425 (  0.858318)
              user     system      total        real
          0.612898   0.247062   0.859960 (  0.861149)
              user     system      total        real
          0.610401   0.244182   0.854583 (  0.855572)

This makes no significant difference than 3
```

---

I used the `Benchmark` tool that's in Ruby by default:

```ruby
# benchmark_pdf.rb
require 'benchmark'
require_relative './lib/format_parser.rb'

n = 1000
Benchmark.bm(7) { |x|
  x.report {
    n.times {
      FormatParser::PDFParser.new.call(File.open('./1_page.pdf', 'rb'))
    }
  }
}

```

### DIFF
--- a/lib/parsers/pdf_parser.rb
+++ b/lib/parsers/pdf_parser.rb
@@ -14,8 +14,8 @@ class FormatParser::PDFParser
   # this. The only way of solving this correctly is by adding
   # different types of PDF's in the specs.
   #
-  COUNT_MARKERS = ['Count ']
-  EOF_MARKER    = '%EOF'
+  MARKERS = { page_count: 'Count ' }
+  MAX_BYTES_TO_READ = 1024
 
   def call(io)
     io = FormatParser::IOConstraint.new(io)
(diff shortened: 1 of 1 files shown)


grdw:
Closing due to performance in large PDF files (> 10MB) is still dreadful

PR #123, opened 2018-06-08
Improve performance of the PDF parser

grdw:
**Changes:**
Uses the [document](https://web.archive.org/web/20141010035745/http://gnupdf.org/Introduction_to_PDF) over here to correctly read and interpret a PDF. Determines the page count correctly for PDF's with:

- Linearized headers
- Catalog headers
- Pages headers

Performance increases significantly.


julik:
Pretty much that - it can read indefinitely (or, I would rather hardcode a number of reads as the maximum) and if it goes to EOF and you use safe_read you will get an exception. That exception will be swallowed by 'FormatParser.parse'. By rwader wrappers I mean the ReadLimiter that gets configured for your parser by the calling method. It ensures you cannot read too much. See the body of 'FormatParser.parse' for how it works ;)

lib/parsers/pdf_parser.rb
```diff
@@ -32,44 +33,57 @@ def call(io)
...
-      case read
-      when '%'
-        break if safe_read(io, EOF_MARKER.size) == EOF_MARKER
-      when '/'
-        find_page_count(io, result)
+    bytes_read = MAX_BYTES_TO_READ
+
+    while read = safe_read(io, bytes_read)
+      if pos = read.index('/')
+        io.seek(io.pos - read.size + pos + 1)
```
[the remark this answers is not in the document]
grdw: This is a good question. Somehow my idea was to read chunks of PDF, and use the `index` of a `/` to set the start position of the next chunk. 

I changed the algorithm a bit to make that concept more clear. Above all it did contain a 'hidden' bug. Namely that the line `Count ` or `Linearized ` were on, always contained some extra `/` which caused `page_count` to be executed for the next chunk nontheless (which might also not happen ... but that somehow didn't happen for the PDF files I tested with). 

However, what *should* happen when the `/` is at the end, is that `find_page_count` returns `nil` and the `io.pos` should move 1 place further than from where it ended in the first run and the loop continues to next chunk.

**To visualize it a little with an example:**

```
# Full string: AA/AAAA/AA/A
# Read in chunks of 3
```

iteration | chunk | io.pos | rewinded pos
---|---|---|---
0 | AA/ | 3 | 3
1 | AAA | 6 | N/A
2 | A/A  | 9 | 8
3 | AA/ | 11 | 11
4 | A | 12 | N/A -> break

grdw:
> See the body of 'FormatParser.parse' for how it works ;)

I see. Makes sense. I can just remove these lines therefor https://github.com/WeTransfer/format_parser/blob/ea8ea3d235a73dbceb1cb83f8eea5dfe2b579f32/lib/parsers/pdf_parser.rb#L54-L57 and write the specs with the `ReadLimiter` included - much like the JPG parser and the TIFF parser?

grdw:
@julik Thanks for the repository over at the `pdf-parser-rework` branch. It helped a lot. I have updated the current implementation and re-used most of your `xref_table` code. I did alter some bits though as some of it didn't make much sense conceptually (like reading from offset to offset). 

I replaced it with something else where which you actually read until the end of a header (`>>`). We could also replace it with a static number of bytes (i.e. 256), but I'd like to hear your opinion on this. 

Also feel free to review it.

lib/parsers/pdf_parser.rb
```diff
@@ -14,62 +15,58 @@ class FormatParser::PDFParser
...
-  EOF_MARKER    = '%EOF'
+  EOF_MARKER = '%%EOF'
 
   def call(io)
     io = FormatParser::IOConstraint.new(io)
 
     return unless safe_read(io, 9) =~ PDF_MARKER
 
-    attributes = scan_for_attributes(io)
+    @xref_table = FormatParser::PDFParser::XRefTable.parse(io)
```
julik: Since we don't have that much extra code does it make sense to have a separate module/file for this?

lib/parsers/pdf_parser.rb
```diff
@@ -14,62 +15,58 @@ class FormatParser::PDFParser
...
+    page_count = nil
 
-    result
-  end
+    @xref_table.each do |xref|
+      io.seek(xref.offset)
+
+      page_count = case safe_read(io, 32)
+                   when /\/Type\s*\/Catalog/
+                     read_header(io, xref, /\/Pages (\d+)/)
```
julik: This one is probably not a very good idea. The `/Pages` element usually contains a reference value (a ref) which is expressed as `<object id> <object generation value> R`. With this regexp you will very likely capture the object ID which has nothing to do with the page count in the document. So you either need to look for something that is guaranteed to contain a literal integer for pagecount (like the `/Count` name) or you need to first figure out what each object from the xref table contains, and _then_ be able to parse (and resolve!) those references. Look at this presentation for example - https://www.adobe.com/technology/pdfs/presentations/KingPDFTutorial.pdf In the "Hello world" even the pages dictionary contains a reference value under `/Pages` name.

In other words, if we do not want to really tokenize and parse the PDF objects with all the related types we got to be super-extra-careful with applying regexes. I'm sorry, I didn't invent this file format 🤷‍♂️

lib/parsers/pdf_parser/xref_table.rb
```diff
@@ -0,0 +1,99 @@
...
+          $3
+        )
+      when /trailer/
+        break
+      end
+    end
+
+    # Check if the number of xrefs we got makes sense
+    if num_objects && num_objects != xref_table.length
+      raise "The xref table was declared to contain #{num_objects}"\
```
julik: This needs a separate exception class, otherwise it will bleed out to the outer level in `FormatParser.parse`. `.parse` swallows all the exceptions that can occur when you exceed read limits, but not the other ones. This is so that we can rapidly find bugs via Appsignal.

lib/parsers/pdf_parser/xref_table.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    @io.seek(xref_table_offset)
+
+    parse_xref_table
+  end
+
+  private
+
+  def xref_table_offset
+    @xref_table_offset ||= begin
```
julik: Does this value need memoization if that value only gets recovered once - and the xref table only gets parsed once per file?

lib/parsers/pdf_parser/xref_table.rb
```diff
@@ -0,0 +1,99 @@
...
+  include FormatParser::IOUtils
+
+  XRef = Struct.new(:idx, :offset, :generation_number, :entry_type)
+
+  def self.parse(io)
+    new(io).parse
+  end
+
+  def initialize(io)
+    @io = io
```
julik: I doubt this object needs to be stateful. It can me a stateless module that we include in the parser itself maybe?

lib/parsers/pdf_parser.rb
julik: That being said, you _can_ use this to find the referenced object in your xref table, seek to it and read the actual '/Pages` object - then you can quickly find the `/Count` which lives inside the `/Pages`. That's what the reference points to.

julik:
I like the idea of reading up to `<<`. It is completely incorrect and very dirty but might just work. In practice `<<` and `>>` define dictionaries which can have arbitrary nesting 😱 and from there on you need a stateful parser, regexes out. *BUT* we believe in dirty. What we _should_ do though is add Measurometer counters for whether the page count could be recovered or not versus just "detected magic header in document" count, so that we can see if our page count detection gets better or worse.

lib/parsers/pdf_parser.rb
grdw: Oh yes. Forgot all about that `/Pages` could also be a reference... 🙄 Yes, than this will break h orribly. Funny that my specs succeed though. 🤔 Will add some more for this specific case though.

lib/parsers/pdf_parser.rb
grdw: It depends, is a 100 lines 'not that much extra code'? Not to sound obtuse or anything, I'm just actually curious about how that lives within WeTransfer. 😅

grdw:
> It is completely incorrect and very dirty but might just work. In practice << and >> define dictionaries which can have arbitrary nesting

Yes, it's either this dirty or reading an arbitrary number of bytes (like 256 for example?) and hoping that the required information is within those number of bytes.

lib/parsers/pdf_parser.rb
julik: Fair fair. Let's leave it the way it has been made now and debate/refactor later ;-)

lib/parsers/pdf_parser.rb
julik: You also want to extend your regexp to include one non-digit for the possible subsequent chars, or at least for a subsequent whitespace or dictionary close (`>>`) to ensure your read didn't truncate the name and the value.

julik:
Yes. Let's _try_ to use it and see how far we get with that approach. On my branch I had a different plan: read the first N bytes of an object using the offset from the xref table, and if those first N bytes seem to contain the chars we are looking for I would actually read the entire object and use a full parser on it. We can always continue on that one if we find the need to.

grdw:
I found a better appraoch than reading dirty *or* reading until `>>`. The actual [end of an object is `endobj`](https://amccormack.net/2012-01-22-anatomy-of-a-pdf-document.html). I created a small module which reads an object until `endobj` and extracts the useful content from it.

grdw:
Side note: the specs don't contain a situation where the `/Catalog` object is needed to count the pages of a PDF. I'm going to add this still though. So that it counts all the `/Type/Page` objects as well.

But I'll add that later on probably. 

Also this still needs the Measurometer.

julik:
If your object contains a `stream` is `endobj` _after_ the `stream` opener or before? If it is after you risk reading the actual JPEGs and other resources too.

grdw:
> If your object contains a stream is endobj after the stream opener or before? If it is after you risk reading the actual JPEGs and other resources too.

It's after the stream 🤔 but I guess we skip those objects by peaking at the first bit of the header [here](https://github.com/WeTransfer/format_parser/pull/123/files#diff-20144cc7370e552d8d96c9f87a809540R46). I believe all `stream...endstream` objects start with a `<</Length>>` header which we don't care that much for.

PR #124, opened 2018-06-11
Finish work on Ogg support

julik:
Continues with  #45

Tested only with Ogg Vorbis. The sample is generated with AWS Polly. The parser fetches:

- audio_sample_rate_hz
- media_duration_seconds
- num_audio_channels

In order to calculate the duration the file has to be sought page by page till
the end.


julik:
Hey @kostyadubinin can you please do a quick check? I don't think I broke anything, but just to be sure.

There are two small changes:

* Specify endianness explicitly for all unpacks
* Get rid of recursive granule search - just do a linear scan per index, back-to-front

Issue #118, opened 2018-05-11
Fixing a bug in id3tag when mapping ID3 tags to intrinsics


julik:
With our fixtures id3tag raises an exception when addressing `Tag#genre` apparently.

```
 1) FormatParser::MP3Parser decodes and estimates duration for a CBR MP3
     Failure/Error: value = tag.public_send(k)
     
     NoMethodError:
       undefined method `gsub' for nil:NilClass
     # /Users/julik/Code/libs/id3tag/lib/id3tag/frames/v2/genre_frame/genre_parser_pre_24.rb:25:in `just_genres'
     # /Users/julik/Code/libs/id3tag/lib/id3tag/frames/v2/genre_frame/genre_parser_pre_24.rb:12:in `genres'
     # /Users/julik/Code/libs/id3tag/lib/id3tag/frames/v2/genre_frame.rb:23:in `get_genres'
     # /Users/julik/Code/libs/id3tag/lib/id3tag/frames/v2/genre_frame.rb:8:in `genres'
```

I've pushed a branch called `id3tag-bug` which demonstrates it. We need to find out why that happens, create a minimal test case and file a PR for https://github.com/krists/id3tag

The `rescue nil` clause in the MP3 parser code can then be removed.

alex-quiterio:
Hello @julik,

After a small investigation around the gem `id3tag` the issue with the call on `nil` seems to be [fixed](https://github.com/krists/id3tag/pull/19) by now. I will wait until a new tag is released in order to make it safe to remove the `rescue nil` from the code. Yaay! 🎉 Always nice to clean up these edge cases from the code.

PR #125, opened 2018-06-13
Remove uneeded guard when ID3Tag tries to find an unknown genre

alex-quiterio:
The guard `rescue nil` can now be safely removed since the original issue
is fixed by https://github.com/krists/id3tag/pull/19

related with: https://github.com/WeTransfer/format_parser/issues/118

### DIFF
--- a/format_parser.gemspec
+++ b/format_parser.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'ks', '~> 0.0.1'
   spec.add_dependency 'exifr', '~> 1.0'
-  spec.add_dependency 'id3tag', '~> 0.10'
+  spec.add_dependency 'id3tag', '~> 0.10', '>= 0.10.1'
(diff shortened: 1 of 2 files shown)


julik:
@alex-quiterio good stuff, thanks! Can you also bump the gem version dependency for id3tag in `.gemspec` since you need it to be `0.10.1` or newer?

julik:
🎉 almost! We probably should use the additional constraint though, since we do not strictly forbid 0.11 and such. See https://guides.rubygems.org/patterns/

> If you want to allow use of newer backwards-compatible versions but need a specific bug fix you can use a compound requirement:

alex-quiterio:
Interesting indeed. I didn't know about the existence of compound requirements. This is actually much better than just adding the constraint in the main version number. Even though, it was a minor bug fix it is always nice to follow the standards on these things. @julik thanks very much for the tip!

julik:
🎉 👍

Issue #93, opened 2018-04-11
We could provide a Railtie to facilitate analysis in ActiveStorage


julik:
As @bermannoah noted ActiveStorage uses MiniMagick (shelling out to `mogrify` and `convert` basically) to detect image formats - https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/analyzer/image_analyzer.rb#L37

We could provide a Railtie that, when triggered in presence of ActiveStorage, would enable FormatParser to perform analysis instead - to the benefit of all involved (shelling out is always more expensive than doing things in-process and FormatParser does a better job in some cases - like CR2 versus TIFFs)

benoittgt:
Hello @julik 

I tried to write a very basic POC for this. I read #126 and tried to write as minimal as possible of new code and use existing API. Also maybe a script like this could be a good idea to try `format_parser` into a very small Rails app. This is base on https://github.com/rails/rails/blob/master/guides/bug_report_templates/active_storage_gem.rb.

I worked on local storage and S3. I tried to let `format_parser` handle as much as possible the treatment of the file for the analyzing. 

If a kind of this code would be implemented, we should provide a guide to inform the user that for an optimal behavior it should provide a `content-type` or disable `identify` when attaching file.
Active Storage does already a basic format parsing with [`Marcel`](https://github.com/basecamp/marcel) in the nominal case:
- https://github.com/rails/rails/blob/faeb1696313b67c4ba4da9fc399a4f1a6a0d61c7/activestorage/app/models/active_storage/blob.rb#L175-L178
- https://github.com/rails/rails/blob/fbe2433be6e052a1acac63c7faf287c52ed3c5ba/activestorage/app/models/active_storage/blob.rb#L259


```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", "6.0.3"
  gem "sqlite3"
  gem "format_parser"
  gem "aws-sdk-s3", require: false
end

require "active_record/railtie"
require "active_storage/engine"
require "tmpdir"

module FormatParser::ActiveStorage
  class BlobAnalyzer < ActiveStorage::Analyzer
    def self.accept?(_blob)
      # We need to stream the file in order to analyze so we canno't say earlier if it match.
      # Our only way is to match content type but FormatParser doesn't seems to do this
      # Other proposal could be to do it with method similar to #metadata but with less info
[...]

PR #126, opened 2018-06-14
Facilitate files analysis for ActiveStorage

codeminator:
Solution for enhancement https://github.com/WeTransfer/format_parser/issues/93

**Assumptions:**

- for `ActiveStorage::Analyzer::ImageAnalyzer`:
  - [rescue block](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/analyzer/image_analyzer.rb#L28) handling **lack of mini_magick gem** has been removed since it's not needed anymore.
  - [private method](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/analyzer/image_analyzer.rb#L41) `#rotated_image?(image)` has been removed:
    - Swapping image height & width if image is rotated is already handled in parsers of `format_parser` gem, for example:
    https://github.com/WeTransfer/format_parser/blob/master/lib/parsers/tiff_parser.rb#L28

**Note:**

Test specs are not there yet because:

- For `require_relative 'railtie' if defined?(Rails)`: If we remove `if defined?(Rails)` part, all existing test specs will fail with the following error:
```
uninitialized constant FormatParser::ActiveStorage::Rails
```

Which makes sense; because we are running the specs, while we are not having a Rails app.

- If we want to add specs for the Railtie, I can see these options:
  - Build a small Rails app inside the spec directory and use it for testing the Railtie (I am in favor of this option since we have only this case ATM).
  - Use this gem https://github.com/pat/combustion, which is achieving the same purpose without the cost of creating small Rails app.
  - Maybe there is no need to test it out, since now we use methods from `format_parser` that are already test-covered.
  - Another option?

@julik  @dennisvdvliet  Please let me know your thoughts.


dennisvdvliet:
@codeminator I would love to see some minimal testing for the integration. So the first option you mentioned (small rails app) seems like the most appropriate.

julik:
TBH I think "monkeypatching by default" is really not a good way to go. It's not the business of a library (any library) to override Rails builtins, however small the override - and especially not for a "utility" library like FormatParser. If ActiveStorage analyzers do not provide sufficient hooks for integrating ourselves as a "third party" analyzer (or making ourselves the default) then these hooks should be made in ActiveStorage itself.

Another concern that I have is that ActiveStorage is cloud-oriented. That means that most of the default backends for ActiveStorage will, one way or another, support `Range:` requests for partial reads from remote - so we can (and should!) exploit `parse_http`. The code here uses the same "download the full monty" approach as the default analyzer, and we are thereby losing the main benefit FormatParser could provide. And do we actually only apply to image analysis?

julik:
> I believe this way, we may use the same Railtie to wrap all this together and add an initializer to set these config values above and add/remove analyzer(s)

Yep, and maybe we shouldn't do this replacement of the builtins in our Railtie but we should make it possible for people to enable our analyzers manually with some setting in `application.rb`. I think we can also substitute `VideoAnalyzer` since we do provide width & height from the MOOV parser for MP4 and MOV (which will be 90% of video files in circulation likely).

codeminator:
@julik Please check the latest commit https://github.com/WeTransfer/format_parser/pull/126/commits/d74daee2d1d0f18a4a1aba11ceaadf719c5d2cb4 :

- It is basically impelementation similar to existing ActiveStorage analyzers (such as [Image Analyzer](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/analyzer/image_analyzer.rb)).
- This is not the final version because:
  - We want to use FormatParser.parse_http(<blob_url>)
  - I looked around and I found the only ways to get the blob URL are:
    - `ActiveStorage::Blob`'s method : `blob.service_url`
    - Rails URL helper method : `rails_blob_url(blob)`
  - Since files analysis in Rails being processed via `ActiveJob` in the background (to avoid blocking the main HTTP request):
    - calling these methods in the background provides the following error:
    ```
    ArgumentError (Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true)
    ``` 
    - It makes sense because the `:host` config option in ActiveStorage disk service [has been eliminated](https://github.com/rails/rails/commit/ccac681122db9747fec9512076772bca345e24b9) (related issue: https://github.com/rails/rails/issues/31842)

  - For demonistration, in `#initialize` method I used `@io = StringIO.new(blob.download)`, this is not the right way to go, as it downloads the whole blob into memory.
  - Then I use `FormatParser.parse(@io)` in `#metadata` method.
  - This works as expected, and analysis is being done correctly via `FormatParser`.

- The solution is to find a way that from inside `FormatParser`, for the `ActiveStorage::Blob` object, is to make `blob.service_url` works.
- I tried several solutions with no luck:
  - Set `ActiveStorage::Current.host` config value in an initializer at Rails app (https://github.com/rails/rails/blob/ec7c89b5160015c014aad0c90f5679252a91cb37/activestorage/app/models/active_storage/current.rb#L4):
    - From console: `blob.service_url` gives the correct url.
    - From a Rails app bundling `FormatParser` gem: gives the same `ArgumentError` stated above.

  - This is a solution (https://github.com/rails/rails/issues/31581#issuecomment-354286466) in order to use these method inside a worker/rake task inside a Rails app, but of course it's not applicable in our case.

  - Currently `ActiveStorage::Current.host` is being set at request level with `request.base_url` (https://github.com/rails/rails/pull/32350/files#diff-b082b6d205256c82b8b6970a21da30a0R8)

Bottom line: If there is a way to get the full url of `ActiveStorage::Blob` object from inside `FormatParser` gem (i.e: get URL from the blob without the need of external Rails help), then we can directly use `FormatParser.parse_http(blob.service_url)` and hence we make the full use of `FormatParser`'s benefits. 

Any though(s) will be appreciated.

julik:
We need to look where this `Blob` thing provides access to a "raw" URL that can be GET'en on, via something like `Aws::S3::Object.presigned_url(:get...` - how do the builtin cloud adapters in ActiveStorage do this?

julik:
Apparently you can get at the `Service`, which in turn supports this: https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/service.rb#L75

The method is `download_chunk` and accepts a key and a range. We can't necessarily use `parse_http` but we can provide a fake IO adapter similar to `RemoteIO` that lives inside FormatParser now to convert `io.seek(offset); io.read(n)` to calls to `Service#download_chunk` - which will probably solve the problem. Why don't we try that? You will need to duplicate/test a bit of the RemoteIO for that to work though.

codeminator:
>  how do the builtin cloud adapters in ActiveStorage do this?
It is being generated in the spot within each service, differs for each one:
- [DiskService](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/service/disk_service.rb#L43)
- [Google Cloud](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/service/gcs_service.rb#L49)
- [S3](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/service/s3_service.rb#L43)

I think that handling this _blob path generation_ in `FormatParser` side will be massive overhead, accordingly, I think that your last suggestion about creating a new IO adapter that "delegates" the read/seek functionality to `blob.service` is the way to go :+1:

codeminator:
@julik I added the Blob IO adapter https://github.com/WeTransfer/format_parser/pull/126/commits/2141525c06798d7aa1293cb8c9e1048c337342d2

Please check and if all good, will proceed with adding more specs and documentation (for IO adapter class and README to illustrate how to hook the new analyzer into a Rails app).

julik:
This is going in the right direction yes, but have you actually tested the entire stack? For example returning a hardcoded `0` from `IO#size` will not get you very far with most parsers. There are also cases with "reading past the end" of the remote resource that need to be covered etc.

codeminator:
@julik for IO adapter I added the size calculation like rest of IO adapters, with more exception handling, and more specs.

Also I updated the README with the how-to enable FormatParser's analyzer

Please check and lemme know your thoughts.

julik:
Yess, good stuff. I will self-assign and review

lib/active_storage/blob_analyzer.rb
```diff
@@ -0,0 +1,30 @@
...
+        true
+      end
+
+      def initialize(blob)
+        @io = BlobIO.new(blob)
+      end
+
+      def metadata
+        result = FormatParser.parse(@io)
+        result.nil? ? {} : result.as_json
```
julik: Do results of FormatParser actually have the same "shape" as, say, the ImageAnalyzer in ActiveStorage? With this you return a Hash that is almost certainly shaped differently.

lib/active_storage/blob_io.rb
```diff
@@ -0,0 +1,54 @@
...
+        range_end =  @pos + n_bytes - 1
+        http_range = (range_start..range_end)
+
+        # ActiveStorageAdapter#download_chunk method return '' when using range (start > end)
+        # Raise invalid range exception if this is the case to recognise the real empty response
+        raise ArgumentError, "Invalid range from #{range_start} to #{range_end}" if range_start > range_end
+
+        begin
+          body = @service.download_chunk(@blob.key, http_range)
+        rescue Errno::ENOENT
```
julik: Are you sure these are the exceptions that the adapters will raise? Do they convert invalid range HTTP code (416 afaik) to EINVAL and ENOENT? Also, the S3 endpoints will give you a 403 (access denied) instead of ENOENT if you try to request from a nonexistent resource unless you happen to have a "List objects" permission on the bucket. How do the service adapters in ActiveStorage deal with these cases?

lib/active_storage/blob_io.rb
```diff
@@ -0,0 +1,54 @@
...
+          raise ArgumentError, "Key #{@blob.key} does not exist"
+        rescue Errno::EINVAL
+          # using negative range for example
+          raise ArgumentError, "Invalid range from #{range_start} to #{range_end}"
+        end
+
+        if body
+          @remote_size = body.bytesize
```
julik: This is very likely not to work. If you do a ranged HTTP request, the returned result will probably indeed be a String and it's `bytesize` will tell you the size of the segment you managed to retrieve. This is also the value that will be in the `Content-Length` response header. It is not the size of the actual resource however ⚠️ - the size of the resource is in the `Content-Range` header returned by the server. The header has the shape of `Content-Range: bytes 0-1023/146515` where the last block of digits is the actual "remote size" you are after.

spec/active_storage/blob_io_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    it 'return the whole file when reading past the end of file' do
+      allow(blob_service).to receive(:download_chunk).with(blob.key, (0..9_999)).and_return('a' * 10)
+
+      io.read(100_00)
+
+      expect(io.size).to eq(10)
+    end
+
+    context 'invalid range' do
+      it 'raise error with -ve range start' do
```
julik: `-ve` does not read as "negative", it reads like l33tspeak. Also, if you have an example with a block that contains nothing RSpec will show it as "green" (passing) whereas in fact the example is a dummy.

lib/active_storage/blob_analyzer.rb
codeminator: That's correct, I considered this in the very first version (Railtie) .. I will try to do it in a way that doesn't break existing attributes usage.

lib/active_storage/blob_io.rb
codeminator: Yes, these exceptions are the ones being raised for DiskStorage, will try to check for [other adapters](https://github.com/rails/rails/tree/master/activestorage/lib/active_storage/service) as well.

lib/active_storage/blob_io.rb
codeminator: I got your point, As I saw that this is the case already within `FormatParser` .. will check how to obtain this header from ActiveStorage response (regardless storage service being used) :+1:

spec/active_storage/blob_io_spec.rb
codeminator: Oops, I forgot to fill this spec!

julik:
Closing for now, feel free to reopen if necessary ;-)

PR #127, opened 2018-06-16
Fixes bytes_used_by_frames variable calculation

krists:
Fixes bytes_used_by_frames variable calculation in FormatParser::MP3Parser class


julik:
❤️ Good one, thanks!

PR #129, opened 2018-06-18
Reset the PDF parser to the utmost basics

grdw:
As discussed today. This PR resets the PDF parser to the basics (in short: removes the dirty page count reader).


julik:
👍

PR #132, opened 2018-08-16
Add :priority when registering parsers and use it when parsing

julik:
Allow the parser priority to be specified. The parsers with a lower priority value will be applied first.
We need this because of two important factors:

* Some parsers are prone to misdetection. For instance, MP3 can be detected in files that aren't really MP3s, just by virtue of having "sort of" a number of MPEG frames. Other parsers, in contrast, are very quick to return early or to detect correctly. Since we rely on "return first" most often, it makes sense to apply the "safer" (most certain) parsers first
* Our file format popularity is roughly this: `JPEG, PDF, PNG, everything else`. It makes sense to apply parsers for the more popular formats first, since a detection is likely to happen earlier.


lorenzograndi4:
@julik 

I added all the files in that transfer to the test suite. All files are recognised as PNGs in both branches, master and detection-order. I havent found a way to reproduce the detection error as we've seen it in the bug report.

julik:
For reference: I've added @lorenzograndi4 's test case file and it indeed did detect as MP3 first, which ended up being returned to the caller as it was the first matching result (and parsers are were registered and then applied in alphabetic order based on the name of the file that contains the parser code). The test cases I've added pertaining to the file indeed to demonstrate incorrect detection on master and correct detection on this branch.

Issue #119, opened 2018-05-23
Add "title", "album" and "artist" to Audio, add M4A/AAC medatada parsing


julik:
For previews it would be very useful to add shortcuts for displaying MP3/MP4 track titles - without having the caller pick them out of the intrinsics attrubute. There is also the unspoken convention that intrinsics can change between minor versions of FormatParser, which can be harmful for long-term storage of metadata. The top-level attributes for audio would be stable, meaning that they would be set regardless.

PR #133, opened 2018-10-02
#119: Add title/album/artist to FormatParser::Audio object

codeminator:
For [Audio struct](https://github.com/WeTransfer/format_parser/blob/master/lib/audio.rb), it will be handy to have direct access to `title`, `artist` and `album` attributes, while fetching them from `id3tags` results (not from `intrinsics` hash), this way, any changes in `intrinsics` would not affect the value of these three attributes.

Note: I found only `Mp3Parser` is the one setting `title`, `album`, `artist` for an audio (inside the `intrinsics` hash)

### DIFF
--- a/lib/audio.rb
+++ b/lib/audio.rb
@@ -4,6 +4,15 @@ module FormatParser
 
     NATURE = :audio
 
+    # Title of the audio
+    attr_accessor :title
+
+    # Album of the audio
+    attr_accessor :album
+
+    # Artist of the audio
+    attr_accessor :artist
+
     # Type of the file (e.g :mp3)
     attr_accessor :format
 
--- a/lib/parsers/mp3_parser.rb
+++ b/lib/parsers/mp3_parser.rb
@@ -81,6 +81,8 @@ class FormatParser::MP3Parser
 
     first_frame = initial_frames.first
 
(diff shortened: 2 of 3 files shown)


codeminator:
@julik Thanks for your feedback 👍 

Please check https://github.com/WeTransfer/format_parser/pull/133/commits/635c660b249582685db6d54e3ef2842d30777d66 :
- Avoid setting attributes with `''`
- In specs depend on actual Mp3 file contains attributes `fixtures_dir + '/MP3/Cassy.mp3'`
- Use `FormatParser.string_to_lossy_utf8` method to ensure string will be valid UTF-8 (please confirm in this part).

lemme know your thoughts.

julik:
My pleasure. The steps are fine. The tests are failing though 🤔  and all in one place https://travis-ci.org/WeTransfer/format_parser/jobs/438477106#L538

codeminator:
@julik Thanks! I think the fix should be trivial .. will do it as soon as I can 👍

codeminator:
@julik I fixed the specs (check last commit), all specs are 💚 Locally, I assume the same for travis-c i as well.

Please let me know if another fix is preferred.

PR #135, opened 2019-07-06
Hint which parser to apply first using the filename

julik:
In some situations we spend way more time applying parsers because the file is in a less-popular format - for example a WAV file. We spend quite a bit of time going through other parsers - some of which are slow. So before we have matched it as a WAV we have to walk through the JPEG parser, then through the ZIP parser (which is quite eager and slow-ish as well), and only at the end we are going to match it as a WAV file.

However, if we know the filename upfront we can do it smarter and apply the parser that is more likely to match first. Note that we won't be assuming the file is of a certain format just based on the filename - we are merely going to rearrange the order of application of the parsers, to optimise it for this particular file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 rvm:
 - 2.2.0
 - 2.6.2
-- jruby-9.0
+- jruby
 sudo: false
 cache: bundler
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 rvm:
 - 2.2.0
-- 2.3.0
-- 2.4.2
-- 2.5.0
+- 2.6.2
 - jruby-9.0
 sudo: false
 cache: bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 rvm:
 - 2.2.0
-- 2.6.2
+- 2.4.6
+- 2.5.5
+- 2.6.3
 - jruby
 sudo: false
 cache: bundler
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -257,7 +257,7 @@ module FormatParser
 
 
     # If there is one parser that is more likely to match, place it first
-    if first_match = factories_in_order_of_priority.find {|f| f.respond_to?(:likely_match?) && f.likely_match?(filename_hint) }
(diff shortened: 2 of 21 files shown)


.travis.yml
```diff
@@ -1,9 +1,7 @@
 rvm:
 - 2.2.0
-- 2.3.0
-- 2.4.2
-- 2.5.0
```
linkyndy: I would keeps supporting all Ruby versions that didn't reach their EOL.

lib/format_parser.rb
```diff
@@ -246,6 +256,12 @@ def self.parsers_for(desired_natures, desired_formats)
       @parser_priorities[parser_factory_a] <=> @parser_priorities[parser_factory_b]
     end
 
+    # If there is one parser that is more likely to match, place it first
+    if first_match = factories_in_order_of_priority.find { |f| f.respond_to?(:likely_match?) && f.likely_match?(filename_hint) }
```
linkyndy: Shouldn't we add `likely_match?` to the parser base class and then (potentially) overwrite it in specific parser classes? This way, we don't have to assert a parser's interface, that it contains a specific method.

.travis.yml
julik: What I wanted to do is to have sufficient "spread" of versions but run less builds. So I've set the lowest supported and the newest supported, assuming that versions in between will kinda work 🕵🏻‍♂️ Do you think it is safe enough, or would you prefer us test with all minor Ruby versions?

lib/format_parser.rb
julik: We could, but here is no base class, on purpose. Even a Proc can be a parser, because it is callable. If we start requiring that all parsers inherit from some arbitrary base class it means that all parsers need to be changed, even ones not stored in this source tree. It also seems odd to require it when all we have is 2 methods to support. The rest is on the parser object to implement (or extend/include if it wishes).

IMO duck-typing in this case is a sensible compromise.

.travis.yml
linkyndy: I would do with all latest minor versions TBH:

```
- 2.4.6
- 2.5.5
- 2.6.3
- jruby
```

lib/format_parser.rb
linkyndy: You're right, I missed that there's no base class. Indeed, that's a "sensible compromise", then.

.travis.yml
julik: Okay, let's do that until we bump major

PR #136, opened 2019-08-03
Improve EXIF handling, for JPEG files in particular

julik:
Since we had a file that was parsed incorrectly, this concludes a little investigation
into our EXIF parsing.

* Fixes the use of the orientation enum
* Adds EXIF flattening for when a file contains multiple EXIF frames


julik:
Interestingly related https://github.com/remvee/exifr/issues/63

lib/parsers/exif_parser.rb
```diff
@@ -65,8 +52,104 @@ def to_json(*maybe_coder)
...
+  # exiftools do this is the way it works.
+  #
+  # This class acts as a wrapper for this "layering" of chunks of EXIF properties, and will
+  # follow the following conventions:
+  #
+  # * When merging data for JSON conversion, it will merge it top-down. It will overwrite
+  #   any specified properties. An exception is made for orientation (see below)
+  # * When retrieving a property, it will look "from the end to the beginning" of the EXIF
+  #   dataframe stack, looking for the first dataframe which has this property with a non-nil value
+  # * When retrieving orientation, it will pick the first orientation value which is not nil
```
dennisvdvliet: If I understand this correctly a JPG can have two (or more) layers of orientation information both telling us something different?

What is reasoning we use different logic for the orientation information vs the other EXIF data?

lib/parsers/jpeg_parser.rb
```diff
@@ -22,6 +22,7 @@ def call(io)
     @width             = nil
     @height            = nil
     @exif_data         = nil
+    @exif_data_frames = []
```
dennisvdvliet: Can you please, please, pretty please add one more space :-)

lib/parsers/jpeg_parser.rb
```diff
@@ -66,16 +67,24 @@ def scan
...
+    # A single file might contain multiple EXIF data frames. In a JPEG this would
+    # manifest as multiple APP1 markers. The way different programs handle these
+    # differs, for us it makes the most sense to simply "flatten" them top-down.
+    # So we start with the first EXIF frame, and we then allow the APP1 markers
+    # that come later in the file to override the properties they _do_ specify.
+    flat_exif = FormatParser::EXIFParser::EXIFStack.new(@exif_data_frames)
+
     # Return at the earliest possible opportunity
     if @width && @height
+      dw, dh = flat_exif.rotated? ? [@height, @width] : [@width, @height]
```
dennisvdvliet: What does `flat_exif.rotated?` return if the images is rotated 180 degrees?

lib/parsers/exif_parser.rb
julik: The reason is that exifr returns the same result (`0`) both when the `0` is recorded in the EXIF tag _and_ when it is simply not present. We need to make a distinction between the two and fallback to the EXIF tag "above" in the latter.

lib/parsers/jpeg_parser.rb
julik: Totally. And remove `@exif_data` which we no longer need

lib/parsers/jpeg_parser.rb
julik: `false`. We could name it something like `orientation_changed?` instead, but I didn't bother because the method is not exposed to the outside from the format parser `Result` object.

PR #142, opened 2019-09-22
Adopt “Hippocratic license”

bermannoah:
format_parser should not be used for evil, and we can update the license to reflect that. (I’d encourage adoption across other OSS projects in this org, but I feel like this is a good place to start.)

For more details: https://firstdonoharm.dev


julik:
I need to discuss this with some people internally, because the license is not set by *us* the developers but by the company.

julik:
@bastiaanterhorst I remember WT OSS release "pact" (when we started publishing code) especially hinges on the agreement that everything we release has to be MIT-licensed. I doubt we could switch to an unproven license like this, even though it has the best intentions. I.e. I think it might cause legal issues  - having a piece of IP with strings attached regarding usage. I leave the matter to you on this.

bermannoah:
If it’s a non-starter I don’t mind closing it.

bastiaanterhorst:
Wowowow not so fast! I am traveling and didn't have a chance to look at this. While in principle I really like what I read, like Julik says, I need to do a legal check.

julik:
I think we should give Bastiaan some time and get back to it if our legal is OK with us doing it ;-) so some patience "wird gebeten"

bastiaanterhorst:
I just got confirmation that we can adopt this license. I suggest we adopt this license as our default from now on.

bermannoah:
Sorry to jump the gun there a little, that is super great news! I've had to recreate the PR in #143.

PR #143, opened 2019-10-23
Adopt Hippocratic License

bermannoah:
format_parser should not be used for evil, and we can update the license to reflect that. (I’d encourage adoption across other OSS projects in this org, but I feel like this is a good place to start.)

For more details: https://firstdonoharm.dev

(Was in #142) 

~Also asking what the license tag in the gemspec should be.~


julik:
Pending reply from Coraline regarding the license tag we need to put in the gemspec.

bermannoah:
@julik I’ve updated the license version and added info to the readme and gemspec. Should we do a major version bump here as well?

julik:
Yes we should!

bermannoah:
@julik 👍 1.0.0? 0.18.0?

julik:
Oof... good question. I think here we are actually at liberty of going 0.18 since we were not at 1.0 to begin with. Let's make use of that fact!

bermannoah:
Beautiful. Updated! :)

Issue #138, opened 2019-09-06
MPG, MPEG support 


grdw:
Currently in Spaceship we have a video thumbnailer. In Spaceship we check the file extension instead of actually using the format parser (mainly because changing the file extension on mobile is quite uncommon). However, we still have plans to use the format parser for this, if only it supported MPG and MPEG (and M4V, but that's probably a bit different so I'll create a different issue for that one). 

I'm wondering if and how this is possible.

Related: https://github.com/WeTransfer/spaceship/issues/1975

PR #149, opened 2020-03-25
Add support for MPEG video files

epintos:
## Description

Adds support for MPEG and MPG video files. Returns the following information:

- format type
- width px
- height px
- aspect ratio
- frame rate

Solves https://github.com/WeTransfer/format_parser/issues/138

## Issues found

Tests run okay, but if I try running an example script, seems like the parser is not properly registered. Might be a file loading issue when testing the gem locally?

## TODOs
- Calculate file size and duration. I wasn't able to parse this properly with the information I found online

## Examples found

I used different test examples from this website: https://standaloneinstaller.com/blog/big-list-of-sample-videos-for-testers-124.html. The license of the website is not clear and uses videos from different sources. Should I get files from another source with a proper license?


lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+    2 => '24',
+    3 => '25',
+    4 => '29.97',
+    5 => '30',
+    6 => '50',
+    7 => '59.94',
+    8 => '60'
+  }
+
+  PACK_HEADER_START_CODE = '000001ba'.freeze
```
julik: You can directly use strings in binary encoding to signify bytes, and compare directly on them. This will free you up from the need to do `to_hex` and back on every comparison / print, but it will also indicate to the reader how many bytes you are planning to compare. If you need a sequence of bytes and it is easy to define it in hexadecimal you can use this:

```
[0x00, 0x00, 0x01, 0xBA].pack("C*")
```

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+class FormatParser::MPEGParser
+  include FormatParser::IOUtils
+
+  MPEG_VALUES = {
+    'mpg' => :mpg,
+    'mpeg' => :mpeg
+  }
+
+  ASPECT_RATIOS = {
+    1 => '1:1',
```
julik: This one is tricky but usually aspects in, for instance, DPX headers are represented as a pair of shorts. What other options can we use for this value that would not have the caller perform string parsing?

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def call(io)
+    return unless matches_mpeg_header?(io)
+
+    # We are looping though the stream because there can be several sequence headers and some of them are not usefull.
+    # If we detect that the header is not usefull, then we look for the next one
+    # If we reach the EOF, then the mpg is likely to be corrupted and we return nil
+    loop do
+      search_for_next_sequence_header(io)
```
julik: `search_` might be somewhat unclear. Do we have better terminology for seeking/skipping? How can we communicate to the reader that this discards the data it reads?

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+      horizontal_size, vertical_size = parse_image_size(io)
+      radio_hex, rate_hex = parse_rate_information(io)
```
julik: `ratio`. Also - what does the `_hex` suffix mean? Does the variable contain a single byte? multiple bytes? a binary string? an integer? But that probably will go away if you remove the to/from hexadecimal conversions ;-)

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      if valid_aspect_radio?(radio_hex) && valid_frame_rate?(rate_hex)
+        return file_info(horizontal_size, vertical_size, radio_hex, rate_hex)
+      end
+    end
+  rescue FormatParser::IOUtils::InvalidRead
+    nil
```
julik: I do not remember exactly but I think this exception is captured by the calling function, can we double-check?

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def file_info(width_px, height_px, radio_hex_raw, rate_hex_raw)
+    FormatParser::Video.new(
+      format: :mpg,
+      width_px: width_px,
+      height_px: height_px,
```
julik: Can we pull framerate up into the main file information struct? The reason is - it seems it will be available always, and we discard the intrinsics when saving metadata because it has a high potential of a) being large and b) including personally identifiable information (EXIF). Same for the aspect ratio. Or will we have to change the struct shape for all video parsers to do so?

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+        aspect_radio: ASPECT_RATIOS.fetch(radio_hex_raw),
+        frame_rate: FRAME_RATES.fetch(rate_hex_raw)
+      },
+    )
+  end
+
+  # The following 3 bytes after the sequence header code, gives us information about the px size
+  # 1.5 bytes (12 bits) for horizontal size and 1.5 bytes for vertical size
+  def parse_image_size(io)
+    image_size = to_hex(safe_read(io, 3))
```
julik: See above re. `to_hex` and the related flow

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+  def matches_mpeg_header?(io)
+    to_hex(safe_read(io, 4)) == PACK_HEADER_START_CODE
+  end
+
+  # Unpacks a whole stream to a unique hexadecimal value
+  def to_hex(value)
+    value.unpack('H*').first
+  end
+
+  def to_decimal(value)
```
julik: `to_decimal` is confusing IMO - because I would suspect it returns something of "a decimal type". If it converts to an integer then it probably could be called `to_i` - but `to_i` is an existing instance method on String. We probably want to explain to the reader why we need a custom method here, and also attempt to explain what `value` is.

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+    value.to_i(16)
+  end
+
+  FormatParser.register_parser new, natures: [:video], formats: MPEG_VALUES.values
```
julik: You can register via `new` but beware that you then must guarantee that your single instance that you create is thread-safe (it can and will be called from multiple threads at the same time). If your parser is stateless it might be easier to specify all the methods as class methods and register `self` instead of `new`

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  ## Searches for the next SEQUENCE_HEADER_START_CODE code, which is the code for a sequence header
+  ## After this code, comes usefull information about the video
+  def search_for_next_sequence_header(io)
+    loop do
+      break if to_hex(safe_read(io, 1)) == SEQUENCE_HEADER_START_CODE
+    end
+  end
+
+  ## If the first 4 bytes of the stream are equal to 00 00 01 BA, the pack start code for the Pack Header, then it's an MPEG file.
```
julik: Let's use a single hashmark when opening a comment please :-)

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def valid_frame_rate?(rate)
+    FRAME_RATES.include?(rate)
+  end
+
+  ## Searches for the next SEQUENCE_HEADER_START_CODE code, which is the code for a sequence header
+  ## After this code, comes usefull information about the video
+  def search_for_next_sequence_header(io)
+    loop do
```
julik: Is there a reasonable upper bound to how many times we want to loop? Since we are dealing with untrusted input it might be a good idea to never do infinite loops, even if there is unbounded read protection in the IO wrapper the parser receives.

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  # The following byte gives us information about the aspect ratio and frame rate
+  # 4 bits corresponds to the aspect ratio and 4 bits to the frame rate code
+  def parse_rate_information(io)
+    rate_information = to_hex(safe_read(io, 1))
+    [rate_information[0].to_i, rate_information[1].to_i]
+  end
+
+  def valid_aspect_radio?(radio)
```
julik: `ratio`. Also the value checked here is not the ratio itself, but an "enum value" - an offset in an arbitrary preselected list of allowed values. We might want to communicate that to the reader.

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+  def parse_image_size(io)
+    image_size = to_hex(safe_read(io, 3))
+    [to_decimal(image_size[0..2]), to_decimal(image_size[3..5])]
+  end
+
+  # The following byte gives us information about the aspect ratio and frame rate
+  # 4 bits corresponds to the aspect ratio and 4 bits to the frame rate code
+  def parse_rate_information(io)
+    rate_information = to_hex(safe_read(io, 1))
+    [rate_information[0].to_i, rate_information[1].to_i]
```
julik: Using raw bytes you can probably use a single `unpack` to split the byte in two sets of 4bits

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+      height_px: height_px,
+      intrinsics: {
+        aspect_radio: ASPECT_RATIOS.fetch(radio_hex_raw),
+        frame_rate: FRAME_RATES.fetch(rate_hex_raw)
+      },
+    )
+  end
+
+  # The following 3 bytes after the sequence header code, gives us information about the px size
+  # 1.5 bytes (12 bits) for horizontal size and 1.5 bytes for vertical size
```
julik: Just out of curiosity - what are the other 12 bits used for?

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
...
+    filename =~ /\.(mpg|mpeg)$/i
+  end
+
+  def call(io)
+    return unless matches_mpeg_header?(io)
+
+    # We are looping though the stream because there can be several sequence headers and some of them are not usefull.
+    # If we detect that the header is not usefull, then we look for the next one
+    # If we reach the EOF, then the mpg is likely to be corrupted and we return nil
+    loop do
```
julik: See below re. unguarded loops

lib/format_parser/version.rb
```diff
@@ -1,3 +1,3 @@
 module FormatParser
-  VERSION = '0.20.1'
+  VERSION = '0.20.2'
```
julik: We might want to bump the minor version instead since we are adding a feature 🚀

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,115 @@
+
+# MPEG Headers documentation:
+# http://dvd.sourceforge.net/dvdinfo/mpeghdrs.html#seq
+# http://www.cs.columbia.edu/~delbert/docs/Dueck%20--%20MPEG-2%20Video%20Transcoding.pdf
+# Useful tool to check the file information: https://www.metadata2go.com/
+class FormatParser::MPEGParser
+  include FormatParser::IOUtils
+
+  MPEG_VALUES = {
```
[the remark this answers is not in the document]
epintos: I'm just using this hash to list the MPEG formats `formats: MPEG_VALUES.values`. I could just refactor it to `formats: [:mpg, :mpeg]`.  Do you think that is easier to read?

lib/parsers/mpeg_parser.rb
epintos: what do you think about `seeking_until_next_sequence_header`?

lib/parsers/mpeg_parser.rb
epintos: I might be getting a bit confused about how unpack,pack works. This is a scenario:
``` ruby
data = safe_read(io, 3) # Returns #\x01@ and I need to read 1.5 bytes for the width and 1.5 bytes for the height 
to_hex(data) # Returns 230140 and then I just want to keep the 230 and the 140 and convert it to decimal (that's why I created to_decimal)
```
For the radio and the rate, I don't need to convert it to a decimal, so I just keep the `hex` value.

Any idea on how to rename these methods properly?

lib/parsers/mpeg_parser.rb
epintos: I thought about this, but if I do that, then all the video parsers will return a null framerate and aspect radio. Do you think that is okay?

lib/parsers/mpeg_parser.rb
epintos: So the header starts with 12 bits for horizontal size, 12 bits for vertical size, 4 bits for aspect ratio, 4 bits for frame rate and 18 bits for the bit rate

lib/parsers/mpeg_parser.rb
epintos: I didn't understand this :/. Can you explain, please?

The aspect ratio has the following format `number : number`, so I think and string is the best way to store that value

lib/parsers/mpeg_parser.rb
epintos: Thanks! That works :)

lib/parsers/mpeg_parser.rb
epintos: M idea was to do that but I struggled a little bit with unpack and pack. Any idea how to do that?

lib/parsers/mpeg_parser.rb
epintos: My first idea was to get that information from the header, but I couldn't find any data that would tell me how many times I need to loop. That is why I added the `rescue FormatParser::IOUtils::InvalidRead`. In case it reaches EOF, it will return nil.

lib/parsers/mpeg_parser.rb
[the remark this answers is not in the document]
julik: I think specifying them inline in the `register_parser` is the best since we are not using those values anywhere else indeed 👍

lib/parsers/mpeg_parser.rb
julik: No problem - aspect ratios are sometimes represented as a pair of `short`s (8-bit unsigned integers) which the caller can the convert to a representation it finds convenient. One system might want to do the calculations in integers, another system might have actual decimal arithmetic available and can prefer that. For example, the DPX format will specify the 1.33333:1 aspect ratio (standard for video / 35mm film) as a tuple of `[4, 3]` which the caller then can divide using arithmetic it finds more convenient. `1.6666666...:1` then becomes `[16, 9]` etc. For aspects which do not have a close integer equivalent you can use `[235, 100]`. Alternatively maybe we can introduce some "minmum viable" FormatParser-wide decimal type?

lib/parsers/mpeg_parser.rb
julik: Close, but with a method named in the "ing-form passive" in Ruby I would expect that this method would be creating / managing some kind of context, continuously (the "how" of some operation). Something like:

```ruby
monitoring_database_load do
   write_ton_of_records
end
```

or

```ruby
capturing_exceptions do
  perform_operation_that_might_fail
end
```

In your case you are imperatively telling the thing to perform a single operation. Maybe `seek_until_sequence_header` or `discard_until_sequence_header` or similar?

lib/parsers/mpeg_parser.rb
epintos: Perfect, I understand now. All the examples I saw online had a `decimal:decimal` format, that's why I was confused. As an MVP version, should we leave this as it is?

lib/parsers/mpeg_parser.rb
julik: Good call. Let's keep it in intrinsics then and postpone the decision 👍

lib/parsers/mpeg_parser.rb
julik: 🤓

lib/parsers/mpeg_parser.rb
julik: As an MVP - ok 👍

lib/parsers/mpeg_parser.rb
julik: We can put a metric here (using Measurometer) to find out "how far" we usually need to loop or how many times. Also your SEQUENCE_HEADER_START_CODE is actually a single byte. You can get better performance if you read (not `safe_read` since it is supposed to guarantee that you have received the amount of bytes you have requested) in larger chunks - say 1KB apiece. Then you can look for the sequence header start code byte inside of it and then `seek` to that place before breaking out of the loop. The reason I am suggesting this is that the IO stack in FormatParser has a limiter on how many `read` calls you perform. If it sees an unreasonable number of `read` calls it assumes the parser has been led into an infinite loop and at some point it will raise an exception. Since all you are doing is looking for a single byte (which you cannot "partially read") reading in bulk and scanning the string you have read will work nicer _and_ keep you under the read count limits.

Re. loop upper bound - if you read by 1024 bytes per iteration, you can for example say that you iterate no more than 1024 times - which will give you an upper bound but also you will know that you are going to be reading at most 1MB of data before giving up.

lib/parsers/mpeg_parser.rb
```diff
@@ -0,0 +1,122 @@
...
+      break if safe_read(io, 1) == SEQUENCE_HEADER_START_CODE
+    end
+  end
+
+  # If the first 4 bytes of the stream are equal to 00 00 01 BA, the pack start code for the Pack Header, then it's an MPEG file.
+  def self.matches_mpeg_header?(io)
+    safe_read(io, 4) == PACK_HEADER_START_CODE
+  end
+
+  def self.convert_3_bytes_to_bits(bytes)
```
julik: ❤️

lib/parsers/mpeg_parser.rb
epintos: Let me know what do you think about the new approach taking this into account :)

PR #153, opened 2020-07-08
Add the option stringify_keys to #as_json

### LINKED ISSUE #151: Return a more uniform structure in as_json
At the moment, results are returned as a regular hash, with keys and values containing a mix of strings and symbols:

```ruby
{"nature"=>:archive, "intrinsics"=>nil, "entries"=>[{:type=>:file, :size=>53335, :filename=>"foo"}, {:type=>:file, :size=>292210, :filename=>"bar"}, {:type=>:file, :size=>1965044, :filename=>"baz"}], "format"=>:zip}
```

It happened to me several times that I was getting the key or comparing the value using the wrong type. I think normalising the return value of `as_json` would be a nice improvement.

Currently, I am doing this to normalise the result hash:

```ruby
JSON.load(JSON.dump(result))
```

Here are the options I see:
* use strings **or** symbols for all key/values (since this is meant to serialise to JSON at _some_ point, it might make sense to stick to strings)
* use a hash with indifferent access; however, that likely adds a dependency and diverts from the idea of having a very strict contract and the least number of dependencies

fabioperrella:
Hi folks! This is my 1st PR in this project.

If I did something wrong, please tell me!!

## Motivation

fix #151

## Proposal

- Add `HashUtils.deep_transform_keys` copied from the activesupport source code

I chose to not add `activesupport` as a dependency because it is good to have the least number of dependencies!

- Use `HashUtils.deep_transform_keys` to transform all the keys of the hash generated by `as_json` to a String

I chose to add the option `stringify_keys` on `#as_json` for backward compatibility. I also added info about it on README

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,17 @@ img_info = FormatParser.parse(File.open("myimage.jpg", "rb"))
 JSON.pretty_generate(img_info) #=> ...
 ```
 
+To convert the result to a Hash or a structure suitable for JSON serialization
+
+```ruby
+img_info = FormatParser.parse(File.open("myimage.jpg", "rb"))
+img_info.as_json
+
+# it's also possible to convert all keys to string
+img_info.as_json(stringify_keys: true)
+```
+
+
 ## Creating your own parsers
 
 See the [section on writing parsers in CONTRIBUTING.md](CONTRIBUTING.md#so-you-want-to-contribute-a-new-parser)
@@ -188,7 +199,7 @@ Unless specified otherwise in this section the fixture files are MIT licensed an
 
 ## Copyright
 
-Copyright (c) 2019 WeTransfer. 
+Copyright (c) 2020 WeTransfer.
 
 `format_parser` is distributed under the conditions of the [Hippocratic License](https://firstdonoharm.dev/version/1/2/license.html)
   - See LICENSE.txt for further details.
--- a/lib/attributes_json.rb
+++ b/lib/attributes_json.rb
@@ -15,7 +15,12 @@ module FormatParser::AttributesJSON
 
   # Implements a sane default `as_json` for an object
   # that accessors defined
-  def as_json(root: false)
+  #
+  # @param root[Bool] if true, it surrounds the result in a hash with a key
+  #  `format_parser_file_info`
+  # @param stringify_keys[Bool] if true, it transforms all the hash keys to a string.
+  #   The default value is false for backward compatibility
+  def as_json(root: false, stringify_keys: false, **)
     h = {}
     h['nature'] = nature if respond_to?(:nature) # Needed for file info structs
     methods.grep(/\w\=$/).each_with_object(h) do |attr_writer_method_name, h|
@@ -27,6 +32,9 @@ module FormatParser::AttributesJSON
       sanitized_value = _sanitize_json_value(unwrapped_attribute_value)
       h[reader_method_name] = sanitized_value
     end
+
+    h = FormatParser::HashUtils.deep_transform_keys(h, &:to_s) if stringify_keys
+
     if root
       {'format_parser_file_info' => h}
     else
--- a/lib/hash_utils.rb
+++ b/lib/hash_utils.rb
@@ -1,0 +1,19 @@
+# based on https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/hash/keys.rb#L116
+# I chose to copy this method instead of adding activesupport as a dependency
+# because we want to have the least number of dependencies
+module FormatParser
+  class HashUtils
+    def self.deep_transform_keys(object, &block)
+      case object
+      when Hash
+        object.each_with_object({}) do |(key, value), result|
+          result[yield(key)] = deep_transform_keys(value, &block)
+        end
+      when Array
+        object.map { |e| deep_transform_keys(e, &block) }
+      else
+        object
+      end
+    end
+  end
+end
--- a/lib/parsers/mp3_parser.rb
+++ b/lib/parsers/mp3_parser.rb
@@ -47,10 +47,6 @@ class FormatParser::MP3Parser
         h[k] = value if value
       end
     end
-
-    def as_json(*)
-      to_h
-    end
   end
 
   def likely_match?(filename)
--- a/spec/attributes_json_spec.rb
+++ b/spec/attributes_json_spec.rb
@@ -140,4 +140,30 @@ describe FormatParser::AttributesJSON do
       JSON.pretty_generate(object_with_attributes_module)
     }.to raise_error(/structure too deep/)
   end
+
+  it 'converts all hash keys to string when stringify_keys: true' do
+    fixture_path = fixtures_dir + '/ZIP/arch_few_entries.zip'
+    fi_io = File.open(fixture_path, 'rb')
+
+    result = FormatParser::ZIPParser.new.call(fi_io).as_json(stringify_keys: true)
+
+    result['entries'].each do |entry|
+      entry.each do |key, _value|
+        expect(key).to be_a(String)
+      end
+    end
+  end
+
+  it 'does not convert hash keys to string when stringify_keys: false' do
+    fixture_path = fixtures_dir + '/ZIP/arch_few_entries.zip'
+    fi_io = File.open(fixture_path, 'rb')
+
+    result = FormatParser::ZIPParser.new.call(fi_io).as_json
+
+    result['entries'].each do |entry|
+      entry.each do |key, _value|
+        expect(key).to be_a(Symbol)
+      end
+    end
+  end
 end
--- a/spec/hash_utils_spec.rb
+++ b/spec/hash_utils_spec.rb
@@ -1,0 +1,42 @@
+require 'spec_helper'
+
+describe FormatParser::HashUtils do
+  describe '.deep_transform_keys' do
+    it 'transforms all the keys in a hash' do
+      hash = { aa: 1, 'bb' => 2 }
+      result = described_class.deep_transform_keys(hash, &:to_s)
+
+      expect(result).to eq('aa' => 1, 'bb' => 2)
+    end
+
+    it 'transforms all the keys in a array of hashes' do
+      array = [{ aa: 1, bb: 2 }, { cc: 3, dd: [{c: 2, d: 3}] }]
+      result = described_class.deep_transform_keys(array, &:to_s)
+
+      expect(result).to eq(
+        [{'aa' => 1, 'bb' => 2}, {'cc' => 3, 'dd' => [{'c' => 2, 'd' => 3}]}]
+      )
+    end
+
+    it 'transforms all the keys in a hash recursively' do
+      hash = { aa: 1, bb: { cc: 22, dd: 3 } }
+      result = described_class.deep_transform_keys(hash, &:to_s)
+
+      expect(result).to eq('aa' => 1, 'bb' => { 'cc' => 22, 'dd' => 3})
+    end
+
+    it 'does nothing for an non array/hash object' do
+      object = Object.new
+      result = described_class.deep_transform_keys(object, &:to_s)
+
+      expect(result).to eq(object)
+    end
+
+    it 'returns the last value if different keys are transformed into the same one' do
+      hash = { aa: 0, 'bb' => 2, bb: 1 }
+      result = described_class.deep_transform_keys(hash, &:to_s)
+
+      expect(result).to eq('aa' => 0, 'bb' => 1)
+    end
+  end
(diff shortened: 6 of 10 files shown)


martijnvermaat:
The Travis tests fail because `pry-byebug` depends on Ruby 2.4. I suggest to separate the `pry-byebug` and `example_status_persistence_file_path` changes into separate PRs so they can be discussed and possibly merged independently.

fabioperrella:
> The Travis tests fail because `pry-byebug` depends on Ruby 2.4. I suggest to separate the `pry-byebug` and `example_status_persistence_file_path` changes into separate PRs so they can be discussed and possibly merged independently.

@martijnvermaat done! I will create another PR with these commits

lib/parsers/mp3_parser.rb
```diff
@@ -47,10 +47,6 @@ def to_h
         h[k] = value if value
       end
     end
-
-    def as_json(*)
-      to_h
-    end
```
martijnvermaat, line 53: It looks like `mp3_parser` had a specialised `as_json` implementation which is now removed?

Perhaps it should instead be something like (pseudo code):
```ruby
def as_json(stringify_keys: false)
  h = to_h
  h = FormatParser::HashUtils.deep_transform_keys(h, &:to_s) if stringify_keys
  h
end
```
Although it shows that the `root` argument was already unsupported by `mp3_parser`...

fabioperrella: This method that I removed was defined in `TagWrapper` which is used only in this piece of code:

```ruby
# lib/parsers/mp3_parser.rb:279
def blend_id3_tags_into_hash(*tags)
  tags.each_with_object({}) do |tag, h|
    h.merge!(TagWrapper.new(tag).to_h)
  end
end
```

Given the fact it returns a `Hash` (and not a `TagWrapper` instance), I'm almost sure that `as_json` is not being used anywhere.

But there is a risk someone is using the class `TagWrapper` directly, but I don't think so, because there is no info about it on README.

Before deleting this method, I ran all the tests to make sure nothing reached there (but I agree this is not a guarantee)

I chose to remove the method instead of using `deep_transform_keys` as you suggested, but I agree it would be a safer approach.

wdyt?

martijnvermaat: I see, thanks for the explanation! I'll defer to someone more familiar with the codebase to comment on that.

README.md
```diff
@@ -188,7 +199,7 @@ Unless specified otherwise in this section the fixture files are MIT licensed an
 
 ## Copyright
 
-Copyright (c) 2019 WeTransfer. 
+Copyright (c) 2019 WeTransfer.
```
linkyndy, line 202: You might as well update the year 😁

lib/hash_utils.rb
```diff
@@ -0,0 +1,19 @@
...
+  class HashUtils
+    def self.deep_transform_keys(object, &block)
+      case object
+      when Hash
+          object.each_with_object({}) do |(key, value), result|
+            result[yield(key)] = deep_transform_keys(value, &block)
+          end
+      when Array
+          object.map { |e| deep_transform_keys(e, &block) }
+        else
```
linkyndy, line 14: *indentation

lib/parsers/mp3_parser.rb
linkyndy: Maybe double check with Julik, see if he has more to add on this.

lib/hash_utils.rb
fabioperrella: I think it messed up when I ran `rubocop -a`, done!

README.md
```diff
@@ -75,6 +75,17 @@ img_info = FormatParser.parse(File.open("myimage.jpg", "rb"))
 JSON.pretty_generate(img_info) #=> ...
 ```
 
+To convert the result to a Hash
```
julik, line 78: If our objective is to convert to a Hash, is the `to_h` method not a better / more intuitive fit?

README.md
```diff
@@ -75,6 +75,17 @@ img_info = FormatParser.parse(File.open("myimage.jpg", "rb"))
...
+
+```ruby
+img_info = FormatParser.parse(File.open("myimage.jpg", "rb"))
+img_info.as_json
+
+# it's also possible to convert all keys to string (since version 0.22)
```
julik, line 84: The README is attached to a specific version of the gem already, so maybe not necessary to mention in which version a certain feature has become available. If it is in the README it means you have it :-)

lib/attributes_json.rb
```diff
@@ -15,7 +15,10 @@ module FormatParser::AttributesJSON
 
   # Implements a sane default `as_json` for an object
   # that accessors defined
-  def as_json(root: false)
+  #
+  # stringify_keys: transforms all the hash keys to a string when is true. The
```
julik, line 19: We have YARD documentation for most methods I think, so the documentation for new parameters should also be in YARD (`@param` and the like)

lib/attributes_json.rb
```diff
@@ -15,7 +15,10 @@ module FormatParser::AttributesJSON
...
+  # default value is false for backward compatibility
+  def as_json(root: false, stringify_keys: false)
```
julik, line 21: Are we sure we support all the options here which ActiveSupport offers? We probably need to add a capture for extra keyword arguments with a `**` for the case that Rails adds another option to `as_json`

lib/hash_utils.rb
```diff
@@ -0,0 +1,19 @@
+# based on https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/hash/keys.rb#L116
+# I chose to copy this method instead of adding activesupport as a dependency
```
julik, line 2: 👍

spec/attributes_json_spec.rb
```diff
@@ -140,4 +140,24 @@ def nature
...
     }.to raise_error(/structure too deep/)
   end
+
+  it 'converts all hash keys to string when stringify_keys: true' do
+    fixture_path = fixtures_dir + '/ZIP/arch_many_entries.zip'
+    fi_io = File.open(fixture_path, 'rb')
+
+    result = FormatParser::ZIPParser.new.call(fi_io).as_json(stringify_keys: true)
+    expect(
+      result['entries'].flat_map { |entry| entry.keys.map(&:class) }.uniq
```
julik, line 150: I would suggest making the expectations a bit less "smart". When dealing with tests it is usually a much more pleasant experience to be visually decoding the code under test, and not the test itself. In this instance the amount of functional sugar used occludes the fact that we want to verify that all keys we are scanning are `kind_of(String)`. Can we use a more pedestrian `each_pair` (or alike) to the same effect?

spec/hash_utils_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'spec_helper'
+
+describe FormatParser::HashUtils do
+  describe '.deep_transform_keys' do
```
julik, line 4: What happens if the hash being transformed has both `"bb"` and `:bb` keys?

lib/parsers/mp3_parser.rb
julik: The idea with this was to ensure that no id3tag gem structures bleed into the FormatParser result, and thus can be serialized to JSON in a simpler/more transparent way. So I think this should work fine.

lib/hash_utils.rb
```diff
@@ -0,0 +1,19 @@
...
+# because we want to have the least number of dependencies
+module FormatParser
+  class HashUtils
```
julik, line 5: Do we need to create an extra module just for this one method? I'm fine with attaching it to the root module. Not critical though :-)

README.md
fabioperrella: I agree on it is a bit confusing! 

Some classes uses to_h like [Array#to_h](https://apidock.com/ruby/v2_5_5/Array/to_h) but other use as_json like [ActiveModel::Serializers::JSON#as_json](https://apidock.com/rails/ActiveModel/Serializers/JSON/as_json)

In this case, the method `as_json` already exists because the classes include the module `FormatParser::AttributesJSON`, so I chose to not change the name.

wdyt?

README.md
fabioperrella: Yeah, I was not sure about adding this information. I will remove it!

lib/attributes_json.rb
fabioperrella: great, I've updated there!

spec/attributes_json_spec.rb
fabioperrella: Sure, I agree the tests should be less "smart"!

Are you suggesting something like this?

```ruby
result['entries'].each do |entry|
  entry.each do |key, value|
    expect(key).to be_a(String)
  end
end
```

I couldn't see a simpler way other than that.

spec/hash_utils_spec.rb
fabioperrella: From what I saw, it will return the lastest value in the hash.

It is the same behavior as active support, and I don't have a better idea for it, do you have?

I added [a spec](https://github.com/WeTransfer/format_parser/blob/a000efa08c2499b5cd329442f2b69566b1fdcfdf/spec/hash_utils_spec.rb#L35-L39) to cover it!

wdyt?

lib/hash_utils.rb
fabioperrella: I understood you are suggesting to add the method `deep_transform_keys` directly in `lib/format_parser.rb`, is it? 🤔 

If it is, I prefer using an extra module and keeping it more in a "single responsibility" way.

wdyt?

lib/attributes_json.rb
fabioperrella: Yeah, we don't support all the options, but should we support it?

I didn't find any issue about it and it has been like this for a long time!

fabioperrella:
tks @julik for all the comments!!

lib/attributes_json.rb
julik: > but should we support it?
Good question - not that we should support them all, but I think we should not crash if they are supplied to us

spec/attributes_json_spec.rb
julik: Yep that's perfect! If you opt for `#each` you might want to group the key-value using `()` though 😉

spec/hash_utils_spec.rb
julik: Fine to just have a spec so that we demonstrate what we expect to happen there 👍

lib/hash_utils.rb
julik: Up to you 👍

julik:
🎩  My pleasure! Will approve once tests are green

lib/attributes_json.rb
fabioperrella: ok, I added the `**`!

fabioperrella:
> 🎩 My pleasure! Will approve once tests are green

I messed up a test! Now that I fixed it, it will be green! :)

spec/attributes_json_spec.rb
fabioperrella: From [what I've seen](https://stackoverflow.com/questions/46435551/in-ruby-whats-the-advantage-of-each-pair-over-each-when-iterating-through-a), `each_pair` is an alias to `each`, isn't it?

>you might want to group the key-value using () though

I didn't get it. Could you give me an example pls?

julik: Sure! I think `some_hash.each do |(key, value)|` is required instead of `some_hash.each do |key, value|` so that the array argument to the block is expanded into two variables. Or you can use `each_pair` instead

README.md
julik: Fair point, maybe change the wording a bit to "To convert the result to a Hash or a structure suitable for JSON serialization" to justify the use of `as_json` a bit better?

spec/attributes_json_spec.rb
fabioperrella: cool! I didn't know about this [parenthesis rule for block variables](https://stackoverflow.com/questions/30326964/rules-on-parenthesis-for-block-variables)!

fabioperrella: Sorry for the long discussion about it, but I found interesting to know how it works and I think it's good to share it with you!

From what I saw, it is not necessary to use the `()` here. It would be necessary if the hash contained arrays as values and I wanted to access their values like this:

image

In this case, there are no arrays inside the hashes, and I just want to check their keys, so it makes no difference using `()`:

image

And also makes no difference using `each` or `each_with_pair` given they are aliases!

image

README.md
fabioperrella: great, much better!

PR #154, opened 2020-07-09
Dev improvements

fabioperrella:
- ~Replaced `pry` by `pry-byebug` for a better debugging, than it's possible to use the commands `step`, `continue`, `up`, `down`, etc.~

- Removed `pry` as a development dependency -> https://github.com/WeTransfer/format_parser/pull/154#discussion_r453323289

- Set up RSpec to use `example_status_persistence_file_path`, so it's possible to run `rspec --only-failures` and `rspec --next-failure` which are pretty cool! (https://relishapp.com/rspec/rspec-core/docs/command-line/only-failures)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ Gemfile.lock
 
 # OSX Files
 .DS_Store
+
+# rspec examples
+spec/examples.txt
--- a/format_parser.gemspec
+++ b/format_parser.gemspec
@@ -39,7 +39,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec', '~> 3.0'
   spec.add_development_dependency 'rake', '~> 12'
   spec.add_development_dependency 'simplecov', '~> 0.15'
-  spec.add_development_dependency 'pry', '~> 0.11'
   spec.add_development_dependency 'yard', '~> 0.9'
   spec.add_development_dependency 'wetransfer_style', '0.5.0'
   spec.add_development_dependency 'parallel_tests'
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -19,6 +19,8 @@ end
 RSpec.configure do |c|
   c.include SpecHelpers
   c.extend SpecHelpers # makes fixtures_dir available for example groups too
+  # https://relishapp.com/rspec/rspec-core/docs/command-line/only-failures
+  c.example_status_persistence_file_path = 'spec/examples.txt'
 end
 
 RSpec.shared_examples 'an IO object compatible with IOConstraint' do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -8,7 +8,6 @@ $LOAD_PATH.unshift(File.dirname(__FILE__))
 
 require 'rspec'
 require 'format_parser'
-require 'pry'
 
 module SpecHelpers
   def fixtures_dir


martijnvermaat:
Added @julik as reviewer since I think his opinion here is more relevant than mine :)

format_parser.gemspec
```diff
@@ -39,7 +39,11 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec', '~> 3.0'
   spec.add_development_dependency 'rake', '~> 12'
   spec.add_development_dependency 'simplecov', '~> 0.15'
-  spec.add_development_dependency 'pry', '~> 0.11'
+  if RUBY_VERSION > '2.4' && RUBY_ENGINE != 'jruby'
```
linkyndy, line 42: This is executed at build time, which is always one of the maintainers' machines. I think you may want to conditionally load these gems at runtime, based on the runtime Ruby.

spec/spec_helper.rb
```diff
@@ -19,6 +19,8 @@ def fixtures_dir
 RSpec.configure do |c|
   c.include SpecHelpers
   c.extend SpecHelpers # makes fixtures_dir available for example groups too
+  # https://relishapp.com/rspec/rspec-core/docs/command-line/only-failures
+  c.example_status_persistence_file_path = 'spec/examples.txt'
```
linkyndy, line 23: I feel this one is kind of in the developers' preference territory. I wouldn't really use it (it's not how I am used to run tests) and I would probably feel "bothered" by that file that always gets added in the dir. I guess others might feel the same.

If you really want this there, I don't mind. If no one has anything against it, feel free to ignore my comment.

fabioperrella: > I feel this one is kind of in the developers' preference territory.

totally agree!

>  would probably feel "bothered" by that file that always gets added in the dir

but even with this file added in .gitignore?

format_parser.gemspec
fabioperrella: > This is executed at build time, which is always one of the maintainers' machines.

yeah!

> I think you may want to conditionally load these gems at runtime, based on the runtime Ruby.

I'm not sure if I understood, but it doesn't matter if using `pry` or `pry-byebug`, you can use the same command as the breakpoint, which is `binding.pry`.

Do you have an example of a problem that could happen to do this way pls?

spec/spec_helper.rb
julik: I'm fine with this change if the file is in gitignore

format_parser.gemspec
julik: Honestly, I think `pry` got into the gemspec by accident and it should be removed alltogether. Rationale:

* A dev dependency must be required to run tests or generate build products. Pry doesn't.
* A dev dependency must be used somewhere in the test/development infrastructure actively. Pry isn't.

The fact that pry merely "is going to be available" can even be considered a net negative, because when working on a gem it will stimulate the developer to add `binding.pry` somewhere and not notice that it actually requires pry to be set as a dependency. Given that I would prefer we _remove_ pry from the gemspec and leave it to the developer to require the variant of pry they prefer :-)

fabioperrella: ok, it makes sense! I've never thought in this way

By doing this, every time I want to use pry, I will need to use it like this?

```ruby
require 'pry' # or 'pry-byebug'
binding.pry
```

(I can edit my snippet to add the "require" line)

julik: Yes, but it will also fail CI build for you and thus remind that there is a stray `binding.pry` somewhere :-)

fabioperrella: humm, that's really a good way to protect us against that!

fabioperrella: So I created [this commit](https://github.com/WeTransfer/format_parser/pull/154/commits/b906535ae842520abce1727f3f07fb2ae879da99) removing `pry` and used some of your words to explain about it!

fabioperrella:
@julik could you check again pls? I think everything is good now

.gitignore
```diff
@@ -58,3 +58,6 @@ Gemfile.lock
 
 # OSX Files
 .DS_Store
+
+# rspec examples
+spec/examples.txt
```
julik, line 63: Add a newline here and it's golden 👍

nitika080289:
> * Replaced `pry` by `pry-byebug` for a better debugging, than it's possible to use the commands `step`, `continue`, `up`, `down`, etc.

Where did we add the pry-byebug dependency?

martijnvermaat:
> Where did we add the pry-byebug dependency?

This change was removed from the PR as per the discussion above ("show resolved").

nitika080289:
> > Where did we add the pry-byebug dependency?
> 
> This change was removed from the PR as per the discussion above ("show resolved").

I see it now. Thanks for pointing it out

fabioperrella:
> > > Where did we add the pry-byebug dependency?
> > 
> > 
> > This change was removed from the PR as per the discussion above ("show resolved").
> 
> I see it now. Thanks for pointing it out

sorry, I updated the description with this!

format_parser.gemspec
linkyndy: Why not specify these kind of dependencies in the `Gemfile`? That's my approach when I don't want to pollute the gemspec with non-build/runtime-related dependencies.

fabioperrella: yeah, it's a good way!

Issue #155, opened 2020-08-13
TypeError in ZIPParser for invalid zips


fabioperrella:
**Message**
```
no implicit conversion of nil into String
```

**Backtrace**
```
/usr/local/bundle/gems/format_parser-0.21.1/lib/parsers/zip_parser/file_reader.rb:178 initialize
/usr/local/bundle/gems/format_parser-0.21.1/lib/parsers/zip_parser/file_reader.rb:178 new
/usr/local/bundle/gems/format_parser-0.21.1/lib/parsers/zip_parser/file_reader.rb:178 read_zip_structure
/usr/local/bundle/gems/format_parser-0.21.1/lib/parsers/zip_parser.rb:17 call
/usr/local/bundle/gems/format_parser-0.21.1/lib/format_parser.rb:201 block in execute_parser_and_capture_expected_exceptions
/usr/local/bundle/gems/measurometer-1.1.0/lib/measurometer.rb:48 instrument
/usr/local/bundle/gems/format_parser-0.21.1/lib/format_parser.rb:200 execute_parser_and_capture_expected_exceptions
/usr/local/bundle/gems/format_parser-0.21.1/lib/format_parser.rb:168 block in parse
/usr/local/bundle/gems/format_parser-0.21.1/lib/format_parser.rb:172 each
/usr/local/bundle/gems/format_parser-0.21.1/lib/format_parser.rb:172 each
```

This issue was being discussed in a private repo and we decided to move the discussion to here.

TLDR:

1. some zip files raise error [at this pont](https://github.com/WeTransfer/format_parser/blob/master/lib/parsers/zip_parser/file_reader.rb#L178) because `io.pos` is beyond the file size 
1. I tested some of those files (in production) with `zip -T` and I got `zip error: Zip file structure invalid`. So I don't think we need to improve the parser to recognize those files
1. I couldn't create a zip file to raise this error to use as a fixture in tests. I tried with `zip -s`, `head -c BYTES file.zip > corrupted.zip`,  etc, but this way, other errors are raised
1. my idea is to create a new error eg: `InvalidCdirPosition = Class.new(Error)` which would be rescued by the `rescue FileReader::Error` (this is the pattern that other errors use), but @linkyndy suggested a flag to raise the errors optionally

@julik replied this:

[...]

PR #156, opened 2020-08-17
Raise a proper error for invalid zips

fabioperrella:
related to https://github.com/WeTransfer/format_parser/issues/155


fabioperrella:
@linkyndy this is the draft PR that I've mentioned about how to deal with the exception raised by some invalid ZIP files

spec/parsers/zip_parser_spec.rb
```diff
@@ -103,4 +103,12 @@
     expect(first_entry.filename).to eq('Li��nia Extreme//')
     expect(first_entry.type).to eq(:directory)
   end
+
+  it 'is able to handle files with invalid central directory position' do
+    invalid_zip_path = fixtures_dir + '/ZIP/invalid_central_directory.zip'
+
+    result = subject.call(File.open(invalid_zip_path, 'rb'))
+
+    expect(result).to be_nil
```
linkyndy, line 112: At first sight, it feels odd that the test for a piece of code that raises an exception returns `nil` 🤔

lib/parsers/zip_parser/file_reader.rb
```diff
@@ -18,6 +18,7 @@ def message
       'Could not find the EOCD signature in the buffer - maybe a malformed ZIP file'
     end
   end
+  CDirInfoNotFound = Class.new(Error)
```
linkyndy, line 21: Do you think we can make the name a little bit more explicit, even if it gets longer? I think it would make exception handling a little bit less cryptic.

fabioperrella: it makes sense! I changed to `InvalidCentralDirectory`

spec/parsers/zip_parser_spec.rb
fabioperrella: you're right! I changed to:

```ruby
expect { subject.call(File.open(invalid_zip_path, 'rb')) }
  .to_not raise_error
```

julik (approved):

julik:
Please apply Rubocop before merging 😉

lib/parsers/zip_parser/file_reader.rb
```diff
@@ -18,6 +18,7 @@ def message
       'Could not find the EOCD signature in the buffer - maybe a malformed ZIP file'
     end
   end
+  InvalidCentralDirectory = Class.new(Error)
```
linkyndy, line 21: Wouldn't it be better to inherit from `StandardError`?

julik: It inherits from `FileReader::Error`, which in turn inherits from `StandardError`. `FileReader::Error` is rescued in the parser modules and returns a `nil`, so that's deliberate.

linkyndy: Oops, my bad then. For some reason I thought it's the top level `Exception` class in Ruby 😅

linkyndy (approved):

nitika080289 (approved):

martijnvermaat (approved):
🎉

PR #162, opened 2020-09-04
MP3 parser: Ignore ID3 tags with empty values

fabioperrella:
When I worked in https://github.com/WeTransfer/format_parser/pull/161 I noticed that the method `TagWrapper#to_h` does not ignore empty strings and because of that, it can replace a id3v1tag by a empty id3v2 tag, so I added a condition to skip for empty values.

Example before this change:
```
v1tag.genre = 'Rock'
v2tag.genre = ''

result: ""
```

Example after this change:
```
v1tag.genre = 'Rock'
v2tag.genre = ''

result: "Rock"
```

The mp3 used as a fixture was created by me. I used the tool `id3ed` (brew install id3ed) to edit the id3v1 tags


julik (approved):
Great find 🔥

nitika080289 (approved):

PR #164, opened 2020-09-08
Tiff parser to return :raw as format for Sony ARW files

### LINKED ISSUE #159: Add support for Sony RAW format images

nitika080289:
Linked to #159


lib/parsers/tiff_parser.rb
```diff
@@ -43,5 +46,9 @@ def cr2?(io)
     safe_read(io, 2) == 'CR'
   end
 
+  def arw?(exif_data)
+    exif_data.compression == 6 && exif_data.new_subfile_type == 1 && exif_data.make == 'SONY'
```
martijnvermaat, line 50: Did you learn this from what exiftool does?

Would be good to mention something about this in a comment, perhaps a link to the equivalent code in exifparser, or to some ARW spec document. Or otherwise a not explaining what `new_subfile_type == 1` means.

lib/parsers/tiff_parser.rb
```diff
@@ -22,6 +22,9 @@ def call(io)
     exif_data = exif_from_tiff_io(io)
     return unless exif_data
 
+    # Skip the parsing if the image is Sony arw format
+    return if arw?(exif_data)
+
     w = exif_data.width || exif_data.pixel_x_dimension
     h = exif_data.height || exif_data.pixel_y_dimension
```
martijnvermaat, line 29: I wonder if we still need the fallbacks if we skip ARWs. In particular because you also removed the test for this 😅  See #146

martijnvermaat (changes requested):
Nice.

I wouldn't say that this "fixes #159" though, at least with its current title 😄  It doesn't add support for ARW files. But I agree that this is a fix for us, at least for now.

lib/parsers/tiff_parser.rb
martijnvermaat: I only read your comment in #159 later. Just having that or something similar as a comment in the code would already be valuable I think.

nitika080289:
> Nice.
> 
> I wouldn't say that this "fixes #159" though, at least with its current title 😄 It doesn't add support for ARW files. But I agree that this is a fix for us, at least for now.

I agree. I think we can keep this PR to solve the issue of handling tiff files with arw format. I will work on a separate PR to add support for files with .arw extension since we may need that later

lib/parsers/tiff_parser.rb
nitika080289: Do you think it would be better to add ARW parser and redirect the parsing to it, to handle this? I am a bit confused because we do a similar thing(ignore parsing) for tiff files with cr2 format.

fabioperrella: @martijnvermaat I didn't get what you mean about the fallbacks..

FormatParser runs [all the parsers](https://github.com/WeTransfer/format_parser/blob/master/lib/format_parser.rb#L170) no matter what is the result of each one.

So, I think there is no way to "disable" the fallbacks.

Maybe the fallback that you mentioned is not the same that I'm thinking of. 🤔

martijnvermaat: @fabioperrella With fallbacks I meant using `pixel_x_dimension` if `width` is not available (same for `y`). I think this is only needed for ARW files, which we are now skipping. Although on closer inspection I see that the original code (before #146 ) used `image_width` so I'm a bit confused.

@nitika080289 I'm not sure what the best way would be to add actual support for ARW. As a separate parser, or to add it to the TIFF parser. But that's something for another PR I'd say, first good step is to stop parsing ARW because we can't really handle it well (for now).

nitika080289: Checked the backend code and looks like these changes should work because now the format parser result would be nil. So the error in #146 would not be raised.

lib/parsers/tiff_parser.rb
linkyndy, line 50: Is this an exhaustive condition? Or there still is the possibility to miss a few ARWs (or the other way around, skip some files that actually are not ARWs)?

nitika080289: To be honest, I am not 100% sure that it will cover all the ARW files or skip some files that are not actually ARWs. I have tested with a few of them(ARW extension, TIFF extension, TIFF extension with ARW format) and it works well. I am relying on the exiftool implementation at the moment. I think we would be able to track the first scenario by looking at the Appsignal errors. Not sure how to tackle the second one though

nitika080289: I have added the exiftool code link since exifr was hard to understand 😆 https://github.com/remvee/exifr/blob/561bbd53b1e1999f60df73ad3e56e9303ed03424/lib/exifr/tiff.rb

martijnvermaat (approved):

fabioperrella (approved):

julik (commented):
I am not sure "removing" ARW support is the way to go to be honest. For example, we do support CR2 files which are also RAW files and cannot be treated like TIFF. Even though it creates more conditionals and branching, for a _user_ of the library having detection work (albeit with not all the useful data supplied) is likely more valuable than not receiving any result at all.

Can we try to solve this a bit more elegantly? (for example - provide a branched detection which would mark ARWs as "raw" and at least try to extract pixel dimensions)? I can help if needed.

julik (approved):
👍

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.24.0
+* Tiff parser to skip Sony RAW format images, to avoid parsing errors
```
julik, line 2: I suggest changing this to "Return :arw from TIFF parser for Sony ARW files"

lib/parsers/tiff_parser.rb
```diff
@@ -22,14 +22,11 @@ def call(io)
...
 
-    # Skip the parsing if the image is Sony arw format
-    return if arw?(exif_data)
-
     w = exif_data.width || exif_data.pixel_x_dimension
     h = exif_data.height || exif_data.pixel_y_dimension
 
     FormatParser::Image.new(
-      format: :tif,
+      format: arw?(exif_data) ? :raw : :tif, # Specify format as raw for Sony RAW format images, else tif
```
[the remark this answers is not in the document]
nitika080289: Modified it to arw. Thanks for pointing it out. I was considering that 'raw' and 'arw' denote the same. But there are other SONY Raw formats like SR2, SRF as well.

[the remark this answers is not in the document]
linkyndy: Yep, as there are RAW formats from other manufacturers as well 😊

linkyndy (approved):

PR #165, opened 2020-09-08
Fix mp3 frames reading to jump correctly to the next bytes

### LINKED ISSUE #163: Does not return metadata for some mp3 files
For some mp3 files, the MP3Parser class doesn't parse the mpeg frames correctly and because of that it fails when tries to find the `frame_bitrate`.

[test.mp3.zip](https://github.com/WeTransfer/format_parser/files/5184863/test.mp3.zip)

This is an example of file with that problem ☝️ . @linkyndy told me that the owner of the file authorized us to use this file as an example, but I'm not sure if it's necessary to mention his name here, is it?

From what I saw, when the parser finds the sync bytes, it's not [getting a valid mpeg_id](https://github.com/WeTransfer/format_parser/blob/603d20c8d01d2d599417821f754e63f3258a183e/lib/parsers/mp3_parser.rb#L209).

These are the expected values according to [the docs](http://www.mpgedit.org/mpgedit/mpeg_format/mpeghdr.htm):
```
# 00 - MPEG Version 2.5
# 01 - reserved
# 10 - MPEG Version 2 (ISO/IEC 13818-3)
# 11 - MPEG Version 1 (ISO/IEC 11172-3)
```

For this file, the first frame that was found returns the value 01, which is `reserved`.

Because of that, it can't find the frame_bitrate and it raises an error `InvalidDeepFetch`.

image

I tested parsing the file with [tinytag](https://pypi.org/project/tinytag/), which the method `parse_mpeg_frames` was based on , and it worked:

```python
from tinytag import TinyTag
tag = TinyTag.get('./test.mp3')
print('bitrate: %s.' % tag.bitrate)
```

```bash
$ python test.py
bitrate: 128
```

fabioperrella:
closes #163 

In MP3Parser, I noticed the [logic to jump to the next bytes](https://github.com/WeTransfer/format_parser/blob/master/lib/parsers/mp3_parser.rb#L147-L150) was wrong (99% sure of that).

To debug it, I added a puts in the method `parse_mpeg_frames`:
```ruby
seek_jmp = sync_bytes_offset_in_4_byte_seq(four_bytes)
puts "frame: #{frame_i}, pos: #{io.pos-4}, pos(h): #{(io.pos-4).to_s(16)} four: #{four_bytes}, hexa: #{four_bytes.map {|ii| ii.to_s(16)} }, jump: #{seek_jmp}"
if seek_jmp > 0
  io.seek(io.pos - 0 + seek_jmp)
  next
end
```

Then I got this:
```
frame: 183, pos: 1474, pos(h): 5c2 four: [0, 255, 251, 148], hexa: ["0", "ff", "fb", "94"], jump: 1
frame: 184, pos: 1479, pos(h): 5c7 four: [0, 0, 0, 0], hexa: ["0", "0", "0", "0"], jump: 4
```

The correct would be jumping to 1475, but it jumped +4 because it didn't take into account the 4 bytes that were already read at the beginning of this method:
```ruby
# lib/parsers/mp3_parser.rb:139
data = io.read(4)
```

I fixed to subtract this 4 bytes, like this:
```ruby
if seek_jmp > 0
  io.seek(io.pos - 4 + seek_jmp)
  next
end
```

And the result was:
```
frame: 366, pos: 1474, pos(h): 5c2 four: [0, 255, 251, 148], hexa: ["0", "ff", "fb", "94"], jump: 1
frame: 367, pos: 1475, pos(h): 5c3 four: [255, 251, 148, 196], hexa: ["ff", "fb", "94", "c4"], jump: 0
```

Because of that, it was possible to read the frames correctly!

I also needed to change MAX_FRAMES_TO_SCAN to a greater value, otherwise, it wouldn't reach this point.

## Attention!

With this modification, the test `#parse_http is called with hash options` in `spec/remote_fetching_spec.rb` started to fail because it started to recognize the file `PNG/anim.png` as an mp3 as well!

I'm not sure if this behaviour can cause new bugs, wdyt?

## Question

~This file was authorized by the owner to use as an example. Should I mention his name in somewhere?`~ (I created a fixture file by myself)


lib/parsers/mp3_parser.rb
```diff
@@ -135,7 +135,8 @@ def parse_mpeg_frames(io)
     MAX_FRAMES_TO_SCAN.times do |frame_i|
       # Read through until we can latch onto the 11 sync bits. Read in 4-byte
       # increments to save on read() calls
-      data = io.read(4)
+      bites_to_read = 4
```
linkyndy, line 138: bytes or bits? 😀

lib/parsers/mp3_parser.rb
```diff
@@ -20,7 +20,7 @@ class InvalidDeepFetch < KeyError
 
   # We limit the number of MPEG frames we scan
   # to obtain our duration estimation
-  MAX_FRAMES_TO_SCAN = 128
+  MAX_FRAMES_TO_SCAN = 500
```
linkyndy, line 23: Just curious, based on what did you get to 500?

lib/parsers/mp3_parser.rb
```diff
@@ -135,7 +135,8 @@ def parse_mpeg_frames(io)
     MAX_FRAMES_TO_SCAN.times do |frame_i|
       # Read through until we can latch onto the 11 sync bits. Read in 4-byte
       # increments to save on read() calls
-      data = io.read(4)
+      bites_to_read = 4
+      data = io.read(bites_to_read)
 
       # If we are at EOF - stop iterating
       break unless data && data.bytesize == 4
```
linkyndy, line 142: Do you want to replace the `4` here as well, with the new variable?

linkyndy (commented):
Nice find!

Though, please remove the file, as mentioned in my DM to you ✌️

lib/parsers/mp3_parser.rb
fabioperrella: good! I took the oportunitty to do this in other places

lib/parsers/mp3_parser.rb
fabioperrella: haha, my bad, it's bytes!

Fixed!

lib/parsers/mp3_parser.rb
fabioperrella: In this file that I used to test, the first frame was found only after the 367th read, so I changed to 500 to give a space for other files which can have that afterwards

Maybe that's something we can improve after, but I'm not sure how yet

wdyt?

fabioperrella:
>Though, please remove the file, as mentioned in my DM to you

I think I can create another file to reproduce the same problem, I will try it!

fabioperrella:
@linkyndy I replaced the file by other one created by myself!

martijnvermaat:
I won't be able to review this until after my holiday, so please proceed without me

lib/parsers/mp3_parser.rb
linkyndy: Is there a standard/guideline on when to expect that first frame within mp3s? 😀  Feels like shooting for the stars here, and it would be nice to make it work for everything, if we can.

I mostly asked because 128 is a _round_ number and 500 isn't 😁

fabioperrella: The doc is here http://www.mpgedit.org/mpgedit/mpeg_format/mpeghdr.htm

>The frame header is constituted by the very first four bytes (32bits) in a frame. The first eleven bits (or first twelve bits, see below about frame sync) of a frame header are always set and they are called "frame sync". Therefore, you can search through the file for the first occurence of frame sync (meaning that you have to find a byte with a value of 255, and followed by a byte with its three (or four) most significant bits set)

From what I understand, the frames don't have a fixed size and we need to find the next "frame sync" to find them.

But, the implementation in format_parser, has a "protection" against huge files, and because of that, it reads every 4 bytes and it has a limit, which was 128 and I increased to 500.

So, I don't know how to define this number more precisely.

Do you have a suggestion?

julik (approved):
Lovely. I have some concerns about the offset arithmetic but let's see how the errors / detection rate look like.

julik:
@fabioperrella I will take a look at the spec

lib/parsers/mp3_parser.rb
linkyndy: Thanks for linking the spec. Let's go with this and see how it behaves in practice 👍

fabioperrella: great!!

linkyndy (approved):

nitika080289 (approved):

PR #169, opened 2020-09-22
MP3: add suport to id3 v2.4.x

### LINKED ISSUE #168: Does not parse mp3 files with id3 v2.4.x
`MP3Parser` skips MP3 files which has id3 version 2.4.x:

```ruby
ID3V2_TAG_VERSIONS = ["\x43\x00".b, "\x03\x00".b, "\x02\x00".b]
# ...
return unless ID3V2_TAG_VERSIONS.include?(header[:version])
```

From what I understand, these versions are 2.2.0, 2.3.0 and 2.67.0

fabioperrella:
fix #168 

## Proposal

Add support for MP3 files with id3 tag version 2.4.x

Refactoring in constant ID3V2_MINOR_TAG_VERSIONS to be easier to understand the values, given the byte notation is a bit hard to understand.

## Comment

I'm not sure why we support the version 2.67 beucase it does not exist. I tried to remove it, but it broke 2 tests. Maybe @julik knows why


spec/parsers/mp3_parser_spec.rb
```diff
@@ -144,6 +144,14 @@
     }.to raise_error(FormatParser::IOUtils::InvalidRead)
   end
 
+  it 'supports id3 v2.4.x' do
+    fpath = fixtures_dir + '/MP3/id3v24.mp3'
```
linkyndy, line 148: What file is this one? If it's the one I sent you, that's private 🙂

lib/parsers/mp3_parser/id3_extraction.rb
```diff
@@ -1,6 +1,9 @@
 module FormatParser::MP3Parser::ID3Extraction
   ID3V1_TAG_SIZE_BYTES = 128
-  ID3V2_TAG_VERSIONS = ["\x43\x00".b, "\x03\x00".b, "\x02\x00".b]
```
linkyndy, line 3: You could've added a comment explaining what these versions are; though I don't mind this change either.

spec/parsers/mp3_parser_spec.rb
fabioperrella: no no! 😄 

it's one that I created by myself!

lib/parsers/mp3_parser/id3_extraction.rb
fabioperrella: yeah, but I think it's a bit easier to understand!

julik (approved):

lib/parsers/mp3_parser/id3_extraction.rb
```diff
@@ -1,6 +1,9 @@
...
+  # it supports 2.4.x, 2.3.x, 2.2.x which are supported by the gem id3tag
+  # see https://id3.org/Developer%20Information for more details of each version
+  # I'm not sure why it has the version 2.67 listed, but it broke 2 tests when I tried to remove it
```
linkyndy, line 5: Do we need to leave this comment here?

fabioperrella: @julik could you help with that pls? do you know the reason of this number 67?

julik:
No idea, need to dive into it still

lib/parsers/mp3_parser/id3_extraction.rb
martijnvermaat: FWIW I could not find any online sources referencing this mythical 2.67 version (also searched for the hexadecimal notation).

It was introduced in format_parser with #115 as a supported version and in an (seemingly, to me) unrelated test without further info on this version number.

So if @julik can't remember, I would suggest removing it (and changing the affected test to some other supported version)? In a separate PR though.

I agree with @linkyndy that this comment should be removed from this PR:
```ruby
# I'm not sure why it has the version 2.67 listed, but it broke 2 tests when I tried to remove it
```

fabioperrella: I opened https://github.com/WeTransfer/format_parser/pull/170 to remove it and rebased this branch with this other, ok?

nitika080289 (approved):

PR #170, opened 2020-09-25
Remove unknown v2.67 from ID3V2_TAG_VERSIONS

fabioperrella:
from https://github.com/WeTransfer/format_parser/pull/169#discussion_r495108090

In MP3Parser, we found out that ID3 v2.4.x is not accepted, but it accepts the version 2.67 which does not exist.

We think this version can be removed.


martijnvermaat (approved):
👍 from me, unless @julik has other thoughts on it of course.

linkyndy (approved):
It would be nice to write a proper description for your PRs; "from comment X" is not very helpful and it usually requires some digging in several places to find the point of the PR—which should be found in its description. You can, of course, also link to the relevant issues if it's useful 😊

fabioperrella:
> It would be nice to write a proper description for your PRs; "from comment X" is not very helpful and it usually requires some digging in several places to find the point of the PR—which should be found in its description. You can, of course, also link to the relevant issues if it's useful 😊

sorry, I added a better description!

julik (approved):

PR #175, opened 2020-09-30
MP3: Fix parse of the Xing header to not raise errors

fabioperrella:
## Motivation

There were some errors in our Appsignal like this:

```
undefined method `unpack' for nil:NilClass

/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.24.2/lib/parsers/mp3_parser.rb:261 attempt_xing_header
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.24.2/lib/parsers/mp3_parser.rb:169 block in parse_mpeg_frames
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.24.2/lib/parsers/mp3_parser.rb:141 times
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.24.2/lib/parsers/mp3_parser.rb:141 parse_mpeg_frames
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.24.2/lib/parsers/mp3_parser.rb:83 call
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.24.2/lib/format_parser.rb:203 block in 
...
```

## Problem 1 - wrong verification for the bitwise operation

I noticed that there were some wrong logics in method `attempt_xing_header` when checking the result of some bitwise operations.

For example: `puts 'true' if anything & 0` will always print `true` because `anything & 0` is equal `0` which is considered true by IF in Ruby.

So, I changed all these operations to check if the result is different than 0.

## Problem 2 - (maybe) wrong unpacking for the Xing header flags

Another thing I'm almost sure was wrong was the directives for unpacking the Xing's `header_flags`.

Before:
ruby
```
# io.read(4).unpack('s>s>')
"\x00\x00\x00\x01".unpack('s>s>')
=> [0, 1]
```

After:
```ruby
# io.read(4).unpack('i>')
"\x00\x00\x00\x01".unpack('i>')
=> [1]
```

The first result would never match the bitwise operations to check if the bits are set.

I based this modification on lib tinytag -> https://github.com/devsnd/tinytag/blob/master/tinytag/tinytag.py#L526

## About the fixture file

To create the fixture file, I downloaded the file from our production server, removed all the frames (in sublimetext hex mode) except the 1st, which has the Xing data and changed the ID3 tags to blur the original file. 

By doing that I think we don't need to ask permission for the owner of the file.


linkyndy (approved):

julik (approved):
Ooh lovely! This should improve MP3 parsing quite a bit, I always wondered why this detection was so broken 😍

fabioperrella:
> Ooh lovely! This should improve MP3 parsing quite a bit, I always wondered why this detection was so broken 😍

nice, I hope so!

PR #176, opened 2020-10-03
Update the contributing doc and documentation

julik:


fabioperrella (approved):
Nice! I didn't know that some parsers were registered with an instance. I thought all of them were registered with classes.

PR #178, opened 2020-10-05
Hotfix Moov parser

### LINKED ISSUE #177: Fix NoMethodError in Moov parser
https://appsignal.com/wetransfer/sites/59550a6bfc940e047fb696bc/exceptions/incidents/644?timerange=2020-10-05T10%3A12%3A00.000Z%2C+2020-10-05T10%3A13%3A00.000Z

lorenzograndi4:
Closes #177 

This is blocking the nu backend release, since there are errors on spaceship which deployed 0.25.1 first

### DIFF
--- a/lib/parsers/moov_parser/decoder.rb
+++ b/lib/parsers/moov_parser/decoder.rb
@@ -36,7 +36,7 @@ class FormatParser::MOOVParser::Decoder
   # matches the type, drilling down if a list of atom names is given
   def find_first_atom_by_path(atoms, *atom_types)
     type_to_find = atom_types.shift
-    requisite = atoms.find { |e| e.is_a?(Atom) && e.atom_type == type_to_find }
+    requisite = atoms.find { |e| e.atom_type == type_to_find }
 
     # Return if we found our match
     return requisite if atom_types.empty?
--- a/lib/parsers/moov_parser/decoder.rb
+++ b/lib/parsers/moov_parser/decoder.rb
@@ -68,7 +68,7 @@ class FormatParser::MOOVParser::Decoder
   def find_video_trak_atom(atoms)
     trak_atoms = find_atoms_by_path(atoms, ['moov', 'trak'])
 
-    return [] if trak_atoms.empty?
+    return if trak_atoms.empty?
 
     trak_atoms.find do |trak_atom|
       hdlr_atom = find_first_atom_by_path([trak_atom], 'trak', 'mdia', 'hdlr')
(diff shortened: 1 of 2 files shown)


lib/parsers/moov_parser/decoder.rb
```diff
@@ -36,7 +36,7 @@ def as_json(*a)
   # matches the type, drilling down if a list of atom names is given
   def find_first_atom_by_path(atoms, *atom_types)
     type_to_find = atom_types.shift
-    requisite = atoms.find { |e| e.atom_type == type_to_find }
+    requisite = atoms.find { |e| e.is_a?(Atom) && e.atom_type == type_to_find }
```
fabioperrella, line 39: I still not understand why the commit 2c8adbf broke it, but this change fix the problem!

fabioperrella (approved):

fabioperrella:
I think I found the real problem..

The method `find_video_trak_atom` sometimes does not find the video trak and it's returning an empty array:

```ruby
def find_video_trak_atom(atoms)
  trak_atoms = find_atoms_by_path(atoms, ['moov', 'trak'])

  return [] if trak_atoms.empty? # <---- here
```

Then, the method `parse_dimensions` assumes the empty array is a found video trak and it shouldn't!

```ruby
def parse_dimensions(decoder, atom_tree)
  video_trak_atom = decoder.find_video_trak_atom(atom_tree)

  tkhd = begin
    if video_trak_atom # <----- here an empty array is considered TRUE
      decoder.find_first_atom_by_path([video_trak_atom], 'trak', 'tkhd')
    else
```

I think the fix should be in method `find_video_trak_atom` to return `nil` when it does not find it:

```ruby
def find_video_trak_atom(atoms)
  trak_atoms = find_atoms_by_path(atoms, ['moov', 'trak'])

  return nil if trak_atoms.empty? # <---- here
```

With that modification, all the tests passed and I tested a file from production having this new issue and it also worked!

wdyt?

fabioperrella (changes requested):
I left a comment before about the problem that I found

lorenzograndi4:
Good catch @fabioperrella :) the empty array played us. 
I'd still leave the class check in the other line, it shouldn't hurt?

fabioperrella (approved):
nice!

fabioperrella:
>I'd still leave the class check in the other line, it shouldn't hurt?

I don't think it's a good choise, it would have hidden this problem, don't you think?

PR #181, opened 2020-12-01
Parse dpx files correctly to avoid FloatDomainError

nitika080289:
For certain dpx format images, in the structure, horizontal_pixel_aspect and vertical_pixel_aspect are zero in the orientation. Since we calculate the pixel_aspect from these values, which in turn is used to determine the display_width_px and display_height_px, the value of pixel_aspect comes up to NaN. This later results in a FloatDomainError

As per my understanding, we should only find the display_width_px and display_height_px when the above values are non-zero, else default these to the width_px and height_px

**Error** NaN
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.22.1/lib/parsers/dpx_parser.rb:49 round
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.22.1/lib/parsers/dpx_parser.rb:49 call
/app/vendor/bundle/ruby/2.7.0/gems/format_parser-0.22.1/lib/format_parser.rb:202 block in execute_parser_and_capture_expected_exceptions

### DIFF
--- a/lib/parsers/dpx_parser.rb
+++ b/lib/parsers/dpx_parser.rb
@@ -35,18 +35,23 @@ class FormatParser::DPXParser
     w = dpx_structure.fetch(:image).fetch(:pixels_per_line)
     h = dpx_structure.fetch(:image).fetch(:lines_per_element)
 
-    pixel_aspect_w = dpx_structure.fetch(:orientation).fetch(:horizontal_pixel_aspect)
-    pixel_aspect_h = dpx_structure.fetch(:orientation).fetch(:vertical_pixel_aspect)
-    pixel_aspect = pixel_aspect_w / pixel_aspect_h.to_f
-
-    image_aspect = w / h.to_f * pixel_aspect
-
     display_w = w
     display_h = h
-    if image_aspect > 1
-      display_h = (display_w / image_aspect).round
-    else
-      display_w = (display_h * image_aspect).round
+
+    pixel_aspect_w = dpx_structure.fetch(:orientation).fetch(:horizontal_pixel_aspect)
+    pixel_aspect_h = dpx_structure.fetch(:orientation).fetch(:vertical_pixel_aspect)
+
+    # Find display height and width based on aspect only if the file structure has pixel aspects
+    if pixel_aspect_h != 0 && pixel_aspect_w != 0
+      pixel_aspect = pixel_aspect_w / pixel_aspect_h.to_f
+
+      image_aspect = w / h.to_f * pixel_aspect
+
+      if image_aspect > 1
+        display_h = (display_w / image_aspect).round
+      else
+        display_w = (display_h * image_aspect).round
+      end
     end
 
     FormatParser::Image.new(


nitika080289:
For creating a sample file with missing horizontal and vertical pixel aspect, I was finding it hard to modify an existing dpx file. Exiftool also does not support this. So, I used an online convertor to convert a sample jpeg file that we have in the fixtures to a dpx. This automatically set the horizontal and vertical aspects to 0, which we wanted to achieve. 
Please let me know if you feel that there can be a better approach to do this.

fabioperrella:
> For creating a sample file with missing horizontal and vertical pixel aspect, I was finding it hard to modify an existing dpx file. Exiftool also does not support this. So, I used an online convertor to convert a sample jpeg file that we have in the fixtures to a dpx. This automatically set the horizontal and vertical aspects to 0, which we wanted to achieve.
> Please let me know if you feel that there can be a better approach to do this.

Nice approach!!

The only problem is that this file is huge (7MB). Is it possible to create a smaller one?

lib/parsers/dpx_parser.rb
```diff
@@ -35,18 +35,23 @@ def call(io)
...
-      display_h = (display_w / image_aspect).round
-    else
-      display_w = (display_h * image_aspect).round
+      image_aspect = w / h.to_f * pixel_aspect
+
+      if image_aspect > 1
+        display_h = (display_w / image_aspect).round
+      else
+        display_w = (display_h * image_aspect).round
+      end
```
fabioperrella, line 54: What about adding a test to cover this fix?

nitika080289:
> > For creating a sample file with missing horizontal and vertical pixel aspect, I was finding it hard to modify an existing dpx file. Exiftool also does not support this. So, I used an online convertor to convert a sample jpeg file that we have in the fixtures to a dpx. This automatically set the horizontal and vertical aspects to 0, which we wanted to achieve.
> > Please let me know if you feel that there can be a better approach to do this.
> 
> Nice approach!!
> 
> The only problem is that this file is huge (7MB). Is it possible to create a smaller one?

That can be done 👍

lib/parsers/dpx_parser.rb
nitika080289: Hmm.. I thought adding the spec would cover the change. I will add a test to verify the values though

fabioperrella: I suggest creating a test to parse this fixture file and to expect that that are no errors in the end.

nitika080289: The first test here does the same https://github.com/WeTransfer/format_parser/blob/master/spec/parsers/dpx_parser_spec.rb

fabioperrella: hmm you're right, this test already does what I suggested!

fabioperrella:
@nitika080289 I noticed that you removed the file, but I think that something must be done to remove this file from git internals, otherwise it will increase the size of the repository.

I'm not sure if a `push --force` would do it, otherwise you can try something like this: https://stackoverflow.com/questions/3458685/how-can-i-completely-remove-a-file-from-a-git-repository

nitika080289:
> @nitika080289 I noticed that you removed the file, but I think that something must be done to remove this file from git internals, otherwise it will increase the size of the repository.
> 
> I'm not sure if a `push --force` would do it, otherwise you can try something like this: https://stackoverflow.com/questions/3458685/how-can-i-completely-remove-a-file-from-a-git-repository
After multiple ways we tried, the file did not actually get deleted from the commits. We decided to leave it like that since it does not have a big impact.

fabioperrella (approved):

martijnvermaat:
If the file is not present in any commit that is reachable from a branch or tag anymore, then it will eventually be garbage-collected by GItHub.

martijnvermaat (approved):

nitika080289:
> If the file is not present in any commit that is reachable from a branch or tag anymore, then it will eventually be garbage-collected by GItHub.

But I see that the file is still present in the first commit in which I added it 8dc54788062d6949a22e756779cf518b51d95e03

martijnvermaat:
Ah right, then it will always be in the history if merging the complete branch. But if you choose "squash and merge" I'm pretty sure it will be gone. In that case your two commits will be squashed into one (that doesn't introduce the large file) and that is the only commit that will later be in the reachable commit graph.

nitika080289:
> Ah right, then it will always be in the history if merging the complete branch. But if you choose "squash and merge" I'm pretty sure it will be gone. In that case your two commits will be squashed into one (that doesn't introduce the large file) and that is the only commit that will later be in the reachable commit graph.

This is nice. It would solve the problem then 👍

julik (approved):

lib/parsers/dpx_parser.rb
```diff
@@ -35,18 +35,23 @@ def call(io)
     w = dpx_structure.fetch(:image).fetch(:pixels_per_line)
     h = dpx_structure.fetch(:image).fetch(:lines_per_element)
 
+    display_w = w
```
linkyndy, line 38: Is there any reason why you moved these lines from their original place? 🤔

nitika080289: I noticed that w and h are already being calculated above and the other statements in between do that impact their values. So the display_w and display_h can be initialised as soon as we have the values of h and w ready.

linkyndy (approved):

PR #182, opened 2020-12-04
Fix MP3Parser to return nil for TIFF files

fabioperrella:
## Motivation

After the commit https://github.com/WeTransfer/format_parser/commit/2c1ddd8c101b0eefd1c5530778d27968bd70e5b8 (version 0.24.1), some TIFF files started to being recognized as MP3 files.

## Proposal

This PR adds a condition in `MP3Parser` to return if it detects that the header of the file includes `TIFF` headers.

Maybe, there is a better approach to solve it, but I don't know how. 

With this solution, there still a chance that `MP3Parser` will recognize other formats as an MP3.

## Fixture file

I used a fixture file created by @linkyndy to create a test for it.


martijnvermaat:
I worry that this will be an endless game of short-circuiting special cases 😅 (Although to be fair that's not where we are yet)

How does format_parser determine the order in which it returns multiple results? A more robust way seems to order them on some kind of confidence level returned by the parser. If that is still a tie, format_parser could prefer "more specific" formats over more generic ones (though I'm not sure what that would mean in this case).

Fact is unfortunately that this is never going to be fully deterministic, because there are many examples of files that are actually valid in different formats.

fabioperrella:
>How does format_parser determine the order in which it returns multiple results?

@martijnvermaat I opened https://github.com/WeTransfer/format_parser/issues/183 which is related to that.

But to explain a little bit, each parser registers itself with a priority, for example [here](https://github.com/WeTransfer/format_parser/blob/master/lib/parsers/mp3_parser.rb#L313). 

But when the priority is the same for more than 1 parser, the order may change (as discussed in the issue above ☝️ )

martijnvermaat:
Ah I see, thanks for explaining and linking to that issue 👍 

Parser priority is indeed part of this (even better if it was deterministic 😃 ), but I can see use cases for ordering results based on how "confident" the parser is as well. But that's opening up a pretty big can of worms so don't want to push for this.

lib/parsers/mp3_parser.rb
```diff
@@ -133,6 +136,10 @@ def call(raw_io)
 
   private
 
+  def tif?(io)
+    FormatParser::TIFFParser::HEADERS.include?(safe_read(io, 4))
```
julik, line 140: If we check against two literal bytebags, I would suggest copying them here instead of creating references between parsers - it is conceptually better, IMO, if the parsers stay independent from one another. Little duplication is likely a better choice here.

fabioperrella: @julik I understand that it would be good to stay independent but on the other side, it seems to me, that it's a responsibility of the `TIFFParser` to know what headers are related to it.

If I duplicate them and they change someday, the MP3Parser would be using the wrong information.

wdyt?

fabioperrella: Ok, got it!

spec/parsers/mp3_parser_spec.rb
```diff
@@ -205,4 +205,12 @@
       ).to eq([ID3Tag::Tag])
     end
   end
+
+  it 'does not recognize TIFF files as MP3' do
+    fpath = fixtures_dir + '/TIFF/tiff_and_mp3.tif'
+
+    parsed = subject.call(File.open(fpath, 'rb'))
+
+    expect(parsed).to eq(nil)
```
linkyndy, line 214: `be_nil` is a bit more readable, and it's also in line with the rest of the file 😇

linkyndy (approved):
Looks good to me.

As a note, I also took that picture from somewhere, though I don't recall from where. Never planned to use it somewhere where I would need the copyright. So it might be better if you found another file with the same "properties".

fabioperrella:
> Looks good to me.
> 
> As a note, I also took that picture from somewhere, though I don't recall from where. Never planned to use it somewhere where I would need the copyright. So it might be better if you found another file with the same "properties".

Luckly 🙏, one of the fixtures files that were already in the repo has the same properties and I removed yours!

julik (approved):

PR #184, opened 2020-12-04
Fix some flaky tests sorting the parsers by name also

### LINKED ISSUE #183: Flaky tests when using results: :first
Some specs, like [this one](https://github.com/WeTransfer/format_parser/blob/master/spec/format_parser_spec.rb#L121-L129), calls the method `FormatParser.parse` with the option `results: :first`, which is the default value, and checks if the returned parser is the expected one.

In some cases, `FormatParser.parse` recognizes a file with multiple formats, and using `results: :first`, it returns only the first one.

The problem is that the order of these results varies.

In [this method](https://github.com/WeTransfer/format_parser/blob/master/lib/format_parser.rb#L257-L259), there is a logic to sort the parsers by priority, but since there are many with the same priority, the order varies on each execution.

```ruby
parsers_in_order_of_priority = parsers.to_a.sort do |parser_a, parser_b|
  @parser_priorities[parser_a] <=> @parser_priorities[parser_b  
end
```

Because of this, some tests fail as the following example:

```
  1) FormatParser.parse when multiple results are requested without :result=> :all (first result) is expected to eq #<FormatParser::Image:0x00007fa3258d28c0 @format=:dpx, @width_px=1, @height_px=1, @display_width_px=1, @display_height_px=1>
     Failure/Error: it { is_expected.to eq(image) }

       expected: #<FormatParser::Image:0x00007fa3258d28c0 @format=:dpx, @width_px=1, @height_px=1, @display_width_px=1, @display_height_px=1>
            got: #<FormatParser::Audio:0x00007fa321af2488 @format=:mp3, @num_audio_channels=2, @audio_sample_rate_hz=48000, @intrinsics={:id3tags=>[]}, @media_duration_seconds=22.06318816438356>
```

fabioperrella:
fix #183 

## Motivation

Some specs, like [this one](https://github.com/WeTransfer/format_parser/blob/master/spec/format_parser_spec.rb#L121-L129), calls the method `FormatParser.parse` with the option `results: :first`, which is the default value, and checks if the returned parser is the expected one.

In some cases, `FormatParser.parse` recognizes a file with multiple formats, and using `results: :first`, it returns only the first one.

The problem is that the order of these results varies.

In [this method](https://github.com/WeTransfer/format_parser/blob/master/lib/format_parser.rb#L257-L259), there is a logic to sort the parsers by priority, but since there are many with the same priority, the order varies on each execution.

```ruby
parsers_in_order_of_priority = parsers.to_a.sort do |parser_a, parser_b|
  @parser_priorities[parser_a] <=> @parser_priorities[parser_b  
end
```

Because of this, some tests fail as the following example:

```
  1) FormatParser.parse when multiple results are requested without :result=> :all (first result) is expected to eq #<FormatParser::Image:0x00007fa3258d28c0 @format=:dpx, @width_px=1, @height_px=1, @display_width_px=1, @display_height_px=1>
     Failure/Error: it { is_expected.to eq(image) }

       expected: #<FormatParser::Image:0x00007fa3258d28c0 @format=:dpx, @width_px=1, @height_px=1, @display_width_px=1, @display_height_px=1>
            got: #<FormatParser::Audio:0x00007fa321af2488 @format=:mp3, @num_audio_channels=2, @audio_sample_rate_hz=48000, @intrinsics={:id3tags=>[]}, @media_duration_seconds=22.06318816438356>
```

I don't know exactly why the order sometimes is different, but I could prove it by adding this line:

```ruby
# lib/format_parser.rb
def self.parsers_for(desired_natures, desired_formats, filename_hint = nil)
  #...
  parsers_in_order_of_priority.each { |parser| puts parser.class } # <--- this one
  parsers_in_order_of_priority
end
```

When I run on my laptop:
```ruby
irb(main):001:0> file = Tempfile.new
irb(main):002:0> FormatParser.parse(file)
FormatParser::GIFParser
Class
FormatParser::PDFParser
FormatParser::PNGParser
FormatParser::MOOVParser
# ...
```

When I run in production:
```ruby
irb(main):005:0> file = Tempfile.new
=> #<Tempfile:/tmp/20201204-5987-1mu9uer>
irb(main):006:0> FormatParser.parse(file)
FormatParser::GIFParser
Class
FormatParser::PNGParser
FormatParser::PDFParser
FormatParser::MOOVParser
# ...
```

I think it's related to the following MUTEX, which sets a few class variables when registering the parsers:
```ruby
# lib/format_parser.rb
PARSER_MUX = Mutex.new

#...

def self.register_parser(callable_parser, formats:, natures:, priority: LEAST_PRIORITY)
  parser_provided_formats = Array(formats)
  parser_provided_natures = Array(natures)
  PARSER_MUX.synchronize do
    @parsers ||= Set.new
    @parsers << callable_parser
```

I also did a test to see if this depend on the order the files are loaded, but the results are the same in staging and also in my laptop:

```ruby
def self.register_parser(callable_parser, formats:, natures:, priority: LEAST_PRIORITY)
  puts "register: #{callable_parser.class}"
```

## Proposal

This PR changes the way the registered parsers are sorted.

It can help us to fix the flaky tests and also getting different results when running `FormatParser.parse` to get only the first result.

### DIFF
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -62,7 +62,9 @@ module FormatParser
         @parsers_per_format[provided_format] << callable_parser
       end
       @parser_priorities ||= {}
+      @parser_order_of_registration ||= []
       @parser_priorities[callable_parser] = priority
+      @parser_order_of_registration << callable_parser
     end
   end
 
@@ -264,9 +266,9 @@ module FormatParser
         # results in different environments, which can be hard to understand why.
         # There is also no guarantee in the order that the elements are added in
         # @@parser_priorities
-        # So, to have always the same order, we sort by the class name when the
-        # priorities are the same.
-        parser_a.class.name <=> parser_b.class.name
+        # So, to have always the same order, we sort by the order that the parsers
+        # were registered if the priorities are the same.
+        @parser_order_of_registration.index(parser_a) <=> @parser_order_of_registration.index(parser_b)
       end
     end
 
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -62,9 +62,9 @@ module FormatParser
         @parsers_per_format[provided_format] << callable_parser
       end
       @parser_priorities ||= {}
-      @parser_order_of_registration ||= []
+      @parser_registration_order ||= []
       @parser_priorities[callable_parser] = priority
-      @parser_order_of_registration << callable_parser
+      @parser_registration_order << callable_parser
     end
   end
 
@@ -268,7 +268,7 @@ module FormatParser
         # @@parser_priorities
         # So, to have always the same order, we sort by the order that the parsers
         # were registered if the priorities are the same.
-        @parser_order_of_registration.index(parser_a) <=> @parser_order_of_registration.index(parser_b)
+        @parser_registration_order.index(parser_a) <=> @parser_registration_order.index(parser_b)
       end
     end
 
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -258,6 +258,14 @@ module FormatParser
       if @parser_priorities[parser_a] != @parser_priorities[parser_b]
         @parser_priorities[parser_a] <=> @parser_priorities[parser_b]
       else
+        # Some parsers have the same priority and we want them to be always sorted
+        # in the same way, to not change the result of FormatParser.parse(results: :first).
+        # When this changes, it can generate flaky tests or event different
+        # results in different environments, which can be hard to understand why.
+        # There is also no guarantee in the order that the elements are added in
+        # @@parser_priorities
+        # So, to have always the same order, we sort by the class name when the
+        # priorities are the same.
         parser_a.class.name <=> parser_b.class.name
       end
     end
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -49,8 +49,8 @@ module FormatParser
     parser_provided_formats = Array(formats)
     parser_provided_natures = Array(natures)
     PARSER_MUX.synchronize do
-      @parsers ||= Set.new
-      @parsers << callable_parser
+      @parsers ||= []
+      @parsers << callable_parser unless @parsers.include?(callable_parser)
       @parsers_per_nature ||= {}
       parser_provided_natures.each do |provided_nature|
         @parsers_per_nature[provided_nature] ||= Set.new
@@ -62,9 +62,7 @@ module FormatParser
         @parsers_per_format[provided_format] << callable_parser
       end
       @parser_priorities ||= {}
-      @parser_registration_order ||= []
       @parser_priorities[callable_parser] = priority
-      @parser_registration_order << callable_parser
     end
   end
 
@@ -268,7 +266,7 @@ module FormatParser
         # @@parser_priorities
         # So, to have always the same order, we sort by the order that the parsers
         # were registered if the priorities are the same.
-        @parser_registration_order.index(parser_a) <=> @parser_registration_order.index(parser_b)
+        @parsers.index(parser_a) <=> @parsers.index(parser_b)
       end
     end
 
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -49,6 +49,8 @@ module FormatParser
     parser_provided_formats = Array(formats)
     parser_provided_natures = Array(natures)
     PARSER_MUX.synchronize do
+      # It can't be a Set because the method `parsers_for` depends on the order
+      # that the parsers were added.
       @parsers ||= []
       @parsers << callable_parser unless @parsers.include?(callable_parser)
       @parsers_per_nature ||= {}
--- a/lib/format_parser.rb
+++ b/lib/format_parser.rb
@@ -255,7 +255,11 @@ module FormatParser
     # Order the parsers according to their priority value. The ones having a lower
     # value will sort higher and will be applied sooner
     parsers_in_order_of_priority = parsers.to_a.sort do |parser_a, parser_b|
-      @parser_priorities[parser_a] <=> @parser_priorities[parser_b]
+      if @parser_priorities[parser_a] != @parser_priorities[parser_b]
+        @parser_priorities[parser_a] <=> @parser_priorities[parser_b]
+      else
+        parser_a.class.name <=> parser_b.class.name
+      end
     end
 
     # If there is one parser that is more likely to match, place it first
--- a/spec/format_parser_spec.rb
+++ b/spec/format_parser_spec.rb
@@ -180,17 +180,17 @@ describe FormatParser do
         [:cr2, :dpx, :fdx, :flac, :gif, :jpg, :mov, :mp4, :m4a, :mp3, :mpg, :mpeg, :ogg, :png, :tif, :wav]
       )
 
-      expect(parsers.map{ |parser| parser.class.name }).to eq([
-        "Class",
-        "FormatParser::GIFParser",
-        "FormatParser::PNGParser",
-        "FormatParser::CR2Parser",
-        "FormatParser::DPXParser",
-        "FormatParser::FLACParser",
-        "FormatParser::MP3Parser",
-        "FormatParser::OggParser",
-        "FormatParser::TIFFParser",
-        "FormatParser::WAVParser"
+      expect(parsers.map { |parser| parser.class.name }).to eq([
+        'Class',
+        'FormatParser::GIFParser',
+        'FormatParser::PNGParser',
+        'FormatParser::CR2Parser',
+        'FormatParser::DPXParser',
+        'FormatParser::FLACParser',
+        'FormatParser::MP3Parser',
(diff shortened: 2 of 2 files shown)


martijnvermaat (approved):

lib/format_parser.rb
```diff
@@ -255,7 +255,11 @@ def self.parsers_for(desired_natures, desired_formats, filename_hint = nil)
     # Order the parsers according to their priority value. The ones having a lower
     # value will sort higher and will be applied sooner
     parsers_in_order_of_priority = parsers.to_a.sort do |parser_a, parser_b|
-      @parser_priorities[parser_a] <=> @parser_priorities[parser_b]
+      if @parser_priorities[parser_a] != @parser_priorities[parser_b]
+        @parser_priorities[parser_a] <=> @parser_priorities[parser_b]
+      else
+        parser_a.class.name <=> parser_b.class.name
```
julik, line 261: We really need to explain _why_ we are doing this in a comment

fabioperrella:
>Also we need to apply autoformatting so that CI passes.

@julik  I tried to reproduce this failure in my laptop but I couldn't!

I'm using rbenv, ruby 2.2.10p489 and rubocop 0.52.1

Do you have any idea of why I can't reproduce it?

image

fabioperrella:
> > Also we need to apply autoformatting so that CI passes.
> 
> @julik I tried to reproduce this failure in my laptop but I couldn't!
> 
> I'm using rbenv, ruby 2.2.10p489 and rubocop 0.52.1
> 
> Do you have any idea of why I can't reproduce it?
> 
> image

Forget about it, I was in the wrong branch as it shows the picture!

fabioperrella:
> I understand why this is useful but I feel like we should explain what we are after with the change. It is not "fix some flaky tests" (which tests? why do they need fixing? are there situations where tests would pass where they should have failed?) but "always apply parsers in the same, deterministic order". Maybe we could find a way to autogenerate the priority value for this, instead of doing a (to me - not very obvious) namesort?
> 
> Also we need to apply autoformatting so that CI passes.

@julik I updated the description with more details. Could you take a look pls?

lib/format_parser.rb
fabioperrella: I added a comment, could you check it pls?

linkyndy:
Thanks, Fabio, for looking into this.

Just my two cents: this issue feels _strange_ especially because it happens on the **same** (single-threaded) environment. On a quick look, the load order for parser files is always the same, same with the parsers array used for the actual matches. Is the [actual sorting](https://github.com/WeTransfer/format_parser/blob/master/lib/format_parser.rb#L257) happening by (string) class name?

fabioperrella:
> Thanks, Fabio, for looking into this.
> 
> Just my two cents: this issue feels _strange_ especially because it happens on the **same** (single-threaded) environment. On a quick look, the load order for parser files is always the same. Is the [actual](https://github.com/WeTransfer/format_parser/blob/master/lib/format_parser.rb#L257) [sorting](https://ruby-doc.org/core-2.5.0/Array.html#method-i-sort) happening by (string) class name?

@linkyndy the current sorting sorts by the priority. Each parser has a priority as follows:
```
@parser_priorities
=> {#<FormatParser::AIFFParser:0x00007ff69d9ee3f0>=>99,
 #<FormatParser::BMPParser:0x00007ff69d9fd080>=>99,
 #<FormatParser::CR2Parser:0x00007ff6a1952d70>=>99,
 #<FormatParser::DPXParser:0x00007ff6a108ef40>=>99,
 #<FormatParser::FDXParser:0x00007ff6a108d028>=>99,
 #<FormatParser::FLACParser:0x00007ff6a097f240>=>99,
 #<FormatParser::GIFParser:0x00007ff69da93f80>=>0,
 FormatParser::JPEGParser=>0,
 #<FormatParser::MOOVParser:0x00007ff6a1900570>=>1,
 #<FormatParser::MP3Parser:0x00007ff6a0951db8>=>99,
 FormatParser::MPEGParser=>99,
 #<FormatParser::OggParser:0x00007ff6a096fac0>=>99,
 #<FormatParser::PDFParser:0x00007ff6a096ce10>=>1,
 #<FormatParser::PNGParser:0x00007ff6a1084ce8>=>1,
 #<FormatParser::PSDParser:0x00007ff6a0977c48>=>99,
 #<FormatParser::TIFFParser:0x00007ff6a0976460>=>99,
 #<FormatParser::WAVParser:0x00007ff69daaba68>=>99,
 #<FormatParser::ZIPParser:0x00007ff6a098c918>=>2}
```

But, when the priority is the same, the method [sort](https://ruby-doc.org/core-2.5.0/Array.html#method-i-sort) does not guarantee that the order of the result will always be the same:

>The result is not guaranteed to be stable. When the comparison of two elements returns 0, the order of the elements is unpredictable.

In fact, I think this explains almost everything!

linkyndy:
> But, when the priority is the same, the method [sort](https://ruby-doc.org/core-2.5.0/Array.html#method-i-sort) does not guarantee that the order of the result will always be the same:
> 
> > The result is not guaranteed to be stable. When the comparison of two elements returns 0, the order of the elements is unpredictable.
> 
> In fact, I think this explains almost everything!

That's where I wanted to get 😊

julik:
Still on the fence about this sorting using a class name. We specify in the documentation for example that a parser can be a Proc. A Proc will not give you a meaningful class name you can sort on, you also pretty much mandate that a parser must be an instance of a fairly unique class :thinking_face: while we can change the library to work this way it does make the API more complicated. How could we solve this differently I wonder?

For example, we add parsers in a certain priority order if a priority value is given. What if we were to assign a better priority value if none is specified, based on the number of parsers already configured?

fabioperrella:
> Still on the fence about this sorting using a class name. We specify in the documentation for example that a parser can be a Proc. A Proc will not give you a meaningful class name you can sort on, you also pretty much mandate that a parser must be an instance of a fairly unique class :thinking_face: while we can change the library to work this way it does make the API more complicated. How could we solve this differently I wonder?
> 
> For example, we add parsers in a certain priority order if a priority value is given. What if we were to assign a better priority value if none is specified, based on the number of parsers already configured?

What if we change the method `register_parser` to sum 0.0001 to the `priority` if an equal priority is already registered?

Example:

```ruby
FormatParser.register_parser(:a, 1)
FormatParser.register_parser(:b, 2)
FormatParser.register_parser(:c, 2)
FormatParser.register_parser(:d, 2)
FormatParser.register_parser(:e, 3)
```

result:
```ruby
[
  [:a, 1], 
  [:b, 2], 
  [:c, 2.0001], 
  [:d, 2.0002], 
  [:e, 3]
]
```

julik:
So the issue that we have is that also parsers with the same priority get sorted undeterministically right?

fabioperrella:
> So the issue that we have is that also parsers with the same priority get sorted undeterministically right?

yes, it is!

julik:
Ok, what if we record the order in which parsers get registered (and when doing a `require` we do it in a deterministic order by doing a `sort` after the glob)? Then we sort by two components - priority and within the priority the order of registration?

fabioperrella:
@julik could you take a look if this is the change that you expected pls?

julik (approved):
🔥  Sweet

lib/format_parser.rb
```diff
@@ -62,7 +62,9 @@ def self.register_parser(callable_parser, formats:, natures:, priority: LEAST_PR
         @parsers_per_format[provided_format] << callable_parser
       end
       @parser_priorities ||= {}
+      @parser_registration_order ||= []
```
linkyndy, line 65: I might be off, but couldn't we reuse `@parsers` by making it an array? It feels redundant to have two structures holding the exact same items, and arrays also provide set methods (wouldn't count performance here, since there're very few registered parsers to begin with).

fabioperrella: Good point!

I changed to be an array, but I'm adding a new element only if it's not there, to act as an Set:
```ruby
@parsers << callable_parser unless @parsers.include?(callable_parser)
```

I think this way will prevent us to introduce a regression.

Is it ok for you?

linkyndy: Sounds reasonable 👍  I'd only check whether there are _really_ no other bits that expect `@parsers` to be a Set.

julik:
I used `Set` initially to make it a no-op if a parser mistakenly gets registered twice. But maybe this was overly cautious, I dunno.

linkyndy:
> I used `Set` initially to make it a no-op if a parser mistakenly gets registered twice. But maybe this was overly cautious, I dunno.

I think it made total sense given ordering was not important. Now that it is, an _ordered set_ would make more sense. But it is good to be cautious with regards to this aspect 🙂

lib/format_parser.rb
fabioperrella: from what I saw, it's ok using an array!

fabioperrella:
@linkyndy I think I replied to your question and now it's ready to be approved. Is there anything else that worries you?

linkyndy (approved):

PR #185, opened 2020-12-07
Add support to ruby 2.7

fabioperrella:
It adds ruby 2.7.2 to travis.yml and it also updates with latest patch version of Ruby 2.5 and 2.6


martijnvermaat (approved):

linkyndy (approved):
Excellent 🙌

As food for thought, should we drop support for very old Ruby versions?

julik (approved):

fabioperrella:
> Excellent 🙌
> 
> As food for thought, should we drop support for very old Ruby versions?

I'm afraid of doing this because it would affect a lot of applications. 

I would do it only if we start having problems to maintain these versions.

But even if we decide to drop them, I would do it in another PR.

Is it ok for you @linkyndy ?

nitika080289 (approved):

linkyndy:
It was just something to think about, no need to act now. But I don't see any reason supporting Ruby versions that are not supported themselves 😊

PR #187, opened 2021-01-06
Support parsing for m3u and m3u8 files

### LINKED ISSUE #186: M3U8 support
Given we're working on features around M3U8 files, it would be great for consistency if format_parser could recognise such files. It's a rather simple text file with only a required header, so it should be fairly trivial to implement.

nitika080289:
Closes #186 

While investigating about the m3u files, I found that there can be three types supported. A plain text file, a file with extended directives and file containing hls m3u extensions. 
https://en.wikipedia.org/wiki/M3U
Files which are extracted playlists from media players fall into second and third category. I think that recognising a plain text file as m3u just on the basis of file extension might not be the right thing to do and there is no other way of recognising it as per my understanding.
The second and third category files contain a header #EXTM3U which can be used to recognise them. This is what I am currently doing here. So, this logic will not recognise a plain text m3u file but will be able to recognise an extended m3u or hls m3u extension.

Also, the directives vary for the second and third category and seems complicated to extract the information from these files. I will create another ticket for extracting more information from the files and work on it separately

### DIFF
--- a/lib/m3u.rb
+++ b/lib/m3u.rb
@@ -1,0 +1,22 @@
+module FormatParser
+  class M3U
+    include FormatParser::AttributesJSON
+
+    NATURE = :text
+
+    attr_accessor :format
+    attr_accessor :size
+
+    # Contains all the file content
+    attr_accessor :content
+
+    # Only permits assignments via defined accessors
+    def initialize(**attributes)
+      attributes.map { |(k, v)| public_send("#{k}=", v) }
+    end
+
+    def nature
+      NATURE
+    end
+  end
+end
--- a/lib/parsers/m3u_parser.rb
+++ b/lib/parsers/m3u_parser.rb
@@ -1,0 +1,25 @@
+class FormatParser::M3UParser
+  include FormatParser::IOUtils
+
+  HEADER = '#EXTM3U'
+
+  def likely_match?(filename)
+    filename =~ /\.(m3u|m3u8)$/i
+  end
+
+  def call(io)
+    io = FormatParser::IOConstraint.new(io)
+
+    header = safe_read(io, 7)
+    return unless HEADER.eql?(header)
+
+    file_content = io.read(io.size)
+
+    FormatParser::M3U.new(
+      format: :m3u,
+      size: io.size,
+      content: file_content
+    )
+  end
+  FormatParser.register_parser new, natures: :text, formats: [:m3u, :m3u8]
+end
--- a/lib/parsers/m3u_parser.rb
+++ b/lib/parsers/m3u_parser.rb
@@ -13,12 +13,9 @@ class FormatParser::M3UParser
     header = safe_read(io, 7)
     return unless HEADER.eql?(header)
 
-    file_content = io.read(io.size)
-
-    FormatParser::M3U.new(
+    FormatParser::Text.new(
       format: :m3u,
-      size: io.size,
-      content: file_content
+      size: io.size
     )
   end
   FormatParser.register_parser new, natures: :text, formats: [:m3u, :m3u8]
--- a/lib/parsers/m3u_parser.rb
+++ b/lib/parsers/m3u_parser.rb
@@ -4,7 +4,7 @@ class FormatParser::M3UParser
   HEADER = '#EXTM3U'
 
   def likely_match?(filename)
-    filename =~ /\.(m3u|m3u8)$/i
+    filename =~ /\.m3u8?$/i
   end
 
   def call(io)
@@ -18,5 +18,5 @@ class FormatParser::M3UParser
       size: io.size
     )
   end
-  FormatParser.register_parser new, natures: :text, formats: [:m3u, :m3u8]
+  FormatParser.register_parser new, natures: :text, formats: :m3u
 end
--- a/lib/parsers/m3u_parser.rb
+++ b/lib/parsers/m3u_parser.rb
@@ -14,8 +14,7 @@ class FormatParser::M3UParser
     return unless HEADER.eql?(header)
 
     FormatParser::Text.new(
-      format: :m3u,
-      size: io.size
+      format: :m3u
     )
   end
   FormatParser.register_parser new, natures: :text, formats: :m3u
--- a/spec/parsers/m3u_parser_spec.rb
+++ b/spec/parsers/m3u_parser_spec.rb
@@ -25,7 +25,6 @@ describe FormatParser::M3UParser do
       expect(parsed_m3u).not_to be_nil
       expect(parsed_m3u.nature).to eq(:text)
       expect(parsed_m3u.format).to eq(:m3u)
-      expect(parsed_m3u.size).to eq(124)
     end
   end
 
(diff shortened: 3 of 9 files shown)


spec/parsers/m3u_parser_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+describe FormatParser::M3UParser do
+  let(:parsed_m3u) {
```
linkyndy, line 4: `do...end` is usually preferred for multi-line blocks 😊

lib/m3u.rb
```diff
@@ -0,0 +1,22 @@
+module FormatParser
+  class M3U
```
linkyndy, line 2: I would name this `Text` instead, to make it in line with the others.

lib/parsers/m3u_parser.rb
```diff
@@ -0,0 +1,25 @@
...
+    filename =~ /\.(m3u|m3u8)$/i
+  end
+
+  def call(io)
+    io = FormatParser::IOConstraint.new(io)
+
+    header = safe_read(io, 7)
+    return unless HEADER.eql?(header)
+
+    file_content = io.read(io.size)
```
linkyndy, line 16: Not sure if we should read the whole content, as the file may be huge and it would defy the purpose of reading as little as possible from the file. Returning the format and size should suffice for now.

linkyndy (commented):
Looks good!

I would say, if we don't need more metadata/directives from the file, we can just ignore them for now.

lib/m3u.rb
nitika080289: I named it M3U considering we will add more "m3u specific" directives here to store information. If we agree that the information we currently have is suffice, I can rename it to Text

lib/parsers/m3u_parser.rb
nitika080289: Right! I will remove this part. Looking at the directives here https://en.wikipedia.org/wiki/M3U, do you think there is any more information we should be parsing?

lib/m3u.rb
fabioperrella: I think he means `M3u` (not using an upper case)

linkyndy: No, I see this in line with `audio`, `video`, `document` etc. M3U is for Text what MP4 is for audio, as I see it.

lib/parsers/m3u_parser.rb
linkyndy: Right now I can't think of anything, no 🤔

lib/parsers/m3u_parser.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    FormatParser::M3U.new(
+      format: :m3u,
+      size: io.size,
+      content: file_content
+    )
+  end
+  FormatParser.register_parser new, natures: :text, formats: [:m3u, :m3u8]
```
fabioperrella, line 24: I'm almost sure that the `natures` must be related to a class which includes `FormatParser::AttributesJSON`, in this case, the `M3u` that you created.

It would be nice to understand if something breaks when using an "invalid" nature and if so, creating a test to cover this. (I think that "text" is an invalid nature)

fabioperrella:
You can also add this new format in `Currently supported filetypes` on Readme!

lib/parsers/m3u_parser.rb
linkyndy: Yes, that's what I meant above with renaming `M3U` to `Text` 🙂

lib/m3u.rb
fabioperrella: hmm, it makes sense!

lib/parsers/m3u_parser.rb
```diff
@@ -0,0 +1,22 @@
+class FormatParser::M3UParser
+  include FormatParser::IOUtils
+
+  HEADER = '#EXTM3U'
+
+  def likely_match?(filename)
+    filename =~ /\.(m3u|m3u8)$/i
```
linkyndy, line 7:
```suggestion
    filename =~ /\.m3u8?$/i
```

This should shorten things a bit.

lib/parsers/m3u_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    header = safe_read(io, 7)
+    return unless HEADER.eql?(header)
+
+    FormatParser::Text.new(
+      format: :m3u,
+      size: io.size
+    )
+  end
+  FormatParser.register_parser new, natures: :text, formats: [:m3u, :m3u8]
```
linkyndy, line 21: Would `:m3u8` ever be returned as a format by format_parser? Looking at jpg/tif, I'd say not 🤔 

It would be interesting to indeed separate m3u from m3u8, but I don't see it as vital.

nitika080289: You are right. It will always return m3u. So m3u8 files will also be recognised as m3u. Actually it might be challenging to distinguish them. I have seen multiple sample files with m3u8 format but they don't have the header to denote the encoding

linkyndy: Indeed, the header is rarely used (from what I've seen), and detecting encodings might be quite a hassle. I'd say this is good for now, as you've implemented it.

lib/parsers/m3u_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def call(io)
+    io = FormatParser::IOConstraint.new(io)
+
+    header = safe_read(io, 7)
+    return unless HEADER.eql?(header)
+
+    FormatParser::Text.new(
+      format: :m3u,
+      size: io.size
```
linkyndy, line 18: Just want to be sure: will this read the whole file in order to get the size? Because if it does, it might be wiser not to include it (or implement it in a different way).

julik: `size` is available once a single read has been performed and will not incur any extra reads

julik (commented):
Everything looks fine - I would be careful with using a property called `size` for a `Text`, the reason is that it is unclear what the unit of `size` is. Is it a byte? (in this case it is). Is it a codepoint? (if the data read was a String in UTF8 and we would call `size` on it it would be). Is it a grapheme cluster? Number of lines? Number of paragraphs?

Also I wonder - do we need to pass the size of the IO to the caller? How is this information going to be used? Maybe we could surface the computed IO size in bytes in all `Result` types if that information is useful and can avoid an extra HTTP request?

nitika080289:
> Everything looks fine - I would be careful with using a property called `size` for a `Text`, the reason is that it is unclear what the unit of `size` is. Is it a byte? (in this case it is). Is it a codepoint? (if the data read was a String in UTF8 and we would call `size` on it it would be). Is it a grapheme cluster? Number of lines? Number of paragraphs?
> 
> Also I wonder - do we need to pass the size of the IO to the caller? How is this information going to be used? Maybe we could surface the computed IO size in bytes in all `Result` types if that information is useful and can avoid an extra HTTP request?

Thanks for your inputs @julik! I just wanted to add some more information about the file along with format. It would not be used by the previewing service but I thought it might be meaningful information for other services using FormatParser outside of our org. If you think this can be removed for now, I will skip it. We can always look into it again if there are any requests to include size.

julik:
@nitika080289 Let's remove it and replace it later with, say, a `bytesize` on all Result types?

linkyndy (approved):
Nice! 🚀

fabioperrella (approved):

julik (approved):

PR #188, opened 2021-01-24
Add MIME types for most popular formats

### LINKED ISSUE #180: Include MIME type when parsing a file
It would useful to include a file's MIME type in its metadata. We would certainly make use of it, as it would already provide a way for clients to consume these files, out-of-the-box (think about the HTML5 `<video>` tag).

julik:
This info will likely be useful in storm, and indeed FormatParser is the best place to emit it.

Closes #180


fabioperrella (approved):
🚀

Issue #189, opened 2021-02-01
Follow redirects when requesting a range in RemoteIO


linkyndy:
Currently, if requesting a range in RemoteIO responds with a 3xx status code, we fail loudly. We should be able to configure the class to use redirects, and optionally the max number of redirects to follow.

PR #190, opened 2021-02-10
Change FormatParser.parse_http to follow a redirect

fabioperrella:
closes https://github.com/WeTransfer/format_parser/issues/189

This PR adds support to follow a redirect when downloading a file with `FormatParser.parse_http`

I was inspired by [this article](https://blog.thecodewhisperer.com/permalink/losing-time-to-faraday) which explains how to follow a redirect using Faraday

### DIFF
--- a/lib/remote_io.rb
+++ b/lib/remote_io.rb
@@ -26,6 +26,7 @@ class FormatParser::RemoteIO
   # @param uri[URI, String] the remote URL to obtain
   def initialize(uri)
     require 'faraday'
+    require 'faraday_middleware/response/follow_redirects'
     @uri = uri
     @pos = 0
     @remote_size = false
@@ -78,7 +79,11 @@ class FormatParser::RemoteIO
     # We use a GET and not a HEAD request followed by a GET because
     # S3 does not allow HEAD requests if you only presigned your URL for GETs, so we
     # combine the first GET of a segment and retrieving the size of the resource
-    response = Faraday.get(@uri, nil, range: 'bytes=%d-%d' % [range.begin, range.end])
+    conn = Faraday.new do |faraday|
+      faraday.use FaradayMiddleware::FollowRedirects
+      faraday.adapter Faraday.default_adapter
+    end
+    response = conn.get(@uri, nil, range: 'bytes=%d-%d' % [range.begin, range.end])
 
     case response.status
     when 200, 206
--- a/lib/remote_io.rb
+++ b/lib/remote_io.rb
@@ -81,6 +81,7 @@ class FormatParser::RemoteIO
     # combine the first GET of a segment and retrieving the size of the resource
     conn = Faraday.new do |faraday|
       faraday.use FaradayMiddleware::FollowRedirects
+      # we still need the default adapter, mode details: https://blog.thecodewhisperer.com/permalink/losing-time-to-faraday
       faraday.adapter Faraday.default_adapter
     end
     response = conn.get(@uri, nil, range: 'bytes=%d-%d' % [range.begin, range.end])
--- a/lib/remote_io.rb
+++ b/lib/remote_io.rb
@@ -81,7 +81,7 @@ class FormatParser::RemoteIO
     # combine the first GET of a segment and retrieving the size of the resource
     conn = Faraday.new do |faraday|
       faraday.use FaradayMiddleware::FollowRedirects
-      # we still need the default adapter, mode details: https://blog.thecodewhisperer.com/permalink/losing-time-to-faraday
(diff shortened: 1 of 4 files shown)


lorenzograndi4 (commented):
LGTM but I'd add a comment (with a link?) to explain why we need default_adapter in the configuration

lorenzograndi4 (approved):
Could we set a limit to the number of redirects to follow?

spec/remote_fetching_spec.rb
```diff
@@ -91,6 +95,13 @@
     end
   end
 
+  context 'when the server respond with a redirect' do
```
linkyndy, line 98:
```suggestion
  context 'when the server responds with a redirect' do
```

😇

spec/remote_io_spec.rb
```diff
@@ -7,7 +7,7 @@
     rio = described_class.new('https://images.invalid/img.jpg')
 
     fake_resp = double(headers: {'Content-Range' => '10-109/2577'}, status: 206, body: 'This is the response')
-    expect(Faraday).to receive(:get).with('https://images.invalid/img.jpg', nil, range: 'bytes=10-109').and_return(fake_resp)
+    expect_any_instance_of(Faraday::Connection).to receive(:get).with('https://images.invalid/img.jpg', nil, range: 'bytes=10-109').and_return(fake_resp)
```
linkyndy, line 10: `expect_any_instance_of` is deprecated for some time now and it is advised against using it. Would it be a better idea to create a stub for `Faraday::Connection` and `expect` on it (similar to what you did below)?

julik (approved):

lib/remote_io.rb
```diff
@@ -78,7 +79,12 @@ def request_range(range)
     # We use a GET and not a HEAD request followed by a GET because
     # S3 does not allow HEAD requests if you only presigned your URL for GETs, so we
     # combine the first GET of a segment and retrieving the size of the resource
-    response = Faraday.get(@uri, nil, range: 'bytes=%d-%d' % [range.begin, range.end])
+    conn = Faraday.new do |faraday|
+      faraday.use FaradayMiddleware::FollowRedirects
+      # we still need the default adapter, mode details: https://blog.thecodewhisperer.com/permalink/losing-time-to-faraday
```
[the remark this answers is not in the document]
fabioperrella: tks!

spec/remote_fetching_spec.rb
fabioperrella: tks again!! 😄

spec/remote_io_spec.rb
fabioperrella: It's weird because RSpec doesn't warn about that in the output, but it is indeed deprecated.

I changed all of them

Thank you!

fabioperrella:
> Could we set a limit to the number of redirects to follow?

@lorenzograndi4 I saw [here](https://www.rubydoc.info/gems/faraday_middleware/0.10.0/FaradayMiddleware%2FFollowRedirects:initialize) that the default value is 3.

It seems a reasonable value for me. 

Do you think we should use another value?

lorenzograndi4:
3 sounds good !

spec/remote_io_spec.rb
```diff
@@ -7,7 +7,9 @@
     rio = described_class.new('https://images.invalid/img.jpg')
 
     fake_resp = double(headers: {'Content-Range' => '10-109/2577'}, status: 206, body: 'This is the response')
-    expect_any_instance_of(Faraday::Connection).to receive(:get).with('https://images.invalid/img.jpg', nil, range: 'bytes=10-109').and_return(fake_resp)
+    faraday_conn = instance_double(Faraday::Connection, get: fake_resp)
```
linkyndy, line 10: You could extract the double in a `let` and the `allow` in a `before`, to DRY things up a bit 😊

fabioperrella: I tried but I think I can't do it.

Faraday is loaded only after calling `FormatParser::RemoteIO.new` ([here](https://github.com/WeTransfer/format_parser/blob/master/lib/remote_io.rb#L28)):
```ruby
def initialize(uri)
  require 'faraday'
```

I think it was done this way to avoid loading the gem when it's not necessary, for example, if you parse only local files.

So, when I tried to reference the Faraday class in the `before` block, I got the error `uninitialized constant Faraday`.

I think that it's worth to require the gem before just to change the test.

Is it ok for you?

linkyndy: Ah, right...

Looking at it better, the subject could also be DRYed out, and that would also require the gem. It is more work than this PR covers, so I'm also fine with moving on with how it is at the moment.

linkyndy (approved):

Issue #191, opened 2021-02-10
Travis takes too much time to run the builds


fabioperrella:
FormatParser uses the free version of Travis to run the builds.

My [last build](https://travis-ci.org/github/WeTransfer/format_parser/builds/758385870) took a very long time to finish it.

We could try to migrate it to GitHub actions.

It's possible to use this project as an example https://github.com/WeTransfer/zip_tricks/blob/main/.github/workflows/ci.yml

PR #192, opened 2021-02-12
Replace Travis by Github actions

fabioperrella:
closes https://github.com/WeTransfer/format_parser/issues/191

It replaces Travis by Github actions.

GH actions runs much faster than the free version of Travis!

It also fixes a problem running tests with JRuby, which was broken in Travis I don't know why.


julik (approved):
🤩

PR #193, opened 2021-02-17
Add option :headers to parse_http

fabioperrella:
related to https://github.com/WeTransfer/format_parser/pull/190

It adds the option `:headers` to `FormatParser.parse_http`, so it's possible to download a file from an URL which requires some kind of authentication header.

### DIFF
--- a/spec/remote_fetching_spec.rb
+++ b/spec/remote_fetching_spec.rb
@@ -114,7 +114,7 @@ describe 'Fetching data from HTTP remotes' do
     it 'uses the provided headers to make the HTTP request' do
       file_information = FormatParser.parse_http(
         'http://localhost:9399//require-auth-and-redirect/TIFF/test.tif',
-       headers: {'Authorization' => 'any-token'}
+        headers: {'Authorization' => 'any-token'}
       )
       expect(file_information.format).to eq(:tif)
     end
--- a/spec/remote_fetching_spec.rb
+++ b/spec/remote_fetching_spec.rb
@@ -19,6 +19,14 @@ describe 'Fetching data from HTTP remotes' do
       res.status = 302
       res.header['Location'] = req.path.sub('/redirect', '')
     end
+    @server.mount_proc '/require-auth-and-redirect' do |req, res|
+      if req.header['authorization'].any?
+        res.status = 302
+        res.header['Location'] = req.path.sub('/require-auth-and-redirect', '')
+      else
+        res.status = 401
+      end
+    end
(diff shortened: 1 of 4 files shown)


spec/remote_fetching_spec.rb
```diff
@@ -102,6 +110,16 @@
     end
   end
 
+  context 'when the server requires authentication' do
```
linkyndy, line 113: I feel this scenario is fairly detailed—authorization is just one use-case for using headers. Wouldn't it be better to generically test whether headers can be passed along with the request?

fabioperrella: I rephrased the test. Do you think that it's enough?

To test that the headers were passed, I would have to stub/mock `Faraday` instance before calling `FormatParser.parse_http`, but it wouldn't work because `Faraday` is required only during execution time.

Do you have other suggestion on how to test it?

julik (approved):

lorenzograndi4 (approved):

spec/remote_fetching_spec.rb
linkyndy: I think naming "authorization" to "test_header" or something similar is already a good step towards abstracting this.

fabioperrella: I think I found a way for that -> 1754f33

Could you take a look pls?

spec/remote_fetching_spec.rb
```diff
@@ -111,9 +103,19 @@
   end
 
   it 'sends provided HTTP headers in the request' do
+    # Faraday is required only after calling .parse_http
+    # This line is just to trigger this require, then it's possible to
+    # add an expectation of how Faraday is initialized after.
+    FormatParser.parse_http('invalid_url') rescue nil
+
+    expect(Faraday)
```
linkyndy, line 111: If you `expect('Faraday')...`, does it still complain about the missing constant?

fabioperrella: it doesn't work! 😞 

image

linkyndy: Bummer. Then, whatever you see fit. The last approach looks very hacky, I think the previous one looked more purposeful, even though it was more involved.

fabioperrella: I prefer the current one because it ensures that all headers were sent.

It's indeed a bit hacky, but I think it's easy to see what is going on reading the comment that I added

ok?

linkyndy (approved):

PR #194, opened 2021-07-30
Upgrade faraday dependency

ljachymczyk:
Faraday gem version used in this project is quite old (from early 2019). It might be nice to upgrade it so that projects using `format_parser` are allowed to benefit from new features of latest `faraday` version.

Checked rspec, it does not seem to break anything. Also tested manually and it seem to work for me.

### DIFF
--- a/format_parser.gemspec
+++ b/format_parser.gemspec
@@ -33,8 +33,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'ks', '~> 0.0'
   spec.add_dependency 'exifr', '~> 1', '>= 1.3.8'
   spec.add_dependency 'id3tag', '~> 0.14'
-  spec.add_dependency 'faraday', '~> 0.13'
-  spec.add_dependency 'faraday_middleware', '~> 0.14'
+  spec.add_dependency 'faraday', '~> 1.5.1'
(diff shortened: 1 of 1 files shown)


format_parser.gemspec
```diff
@@ -33,8 +33,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'ks', '~> 0.0'
   spec.add_dependency 'exifr', '~> 1', '>= 1.3.8'
   spec.add_dependency 'id3tag', '~> 0.14'
-  spec.add_dependency 'faraday', '~> 0.13'
-  spec.add_dependency 'faraday_middleware', '~> 0.14'
+  spec.add_dependency 'faraday', '~> 1.5.1'
```
fabioperrella, line 36:
```suggestion
  spec.add_dependency 'faraday', '~> 1.5'
```

What about being more flexible about it? Theoretically, minor version bumps don't break compatibility

fabioperrella:
Thank you @ljachymczyk for your contribution!

fabioperrella:
I see that the tests are failing for old Rubies.

What about dropping support for EOL Rubies?

```
# format_parser.gemspec
spec.required_ruby_version = Gem::Requirement.new(">= 2.6.0")
```

Also changing the matrix on `format_parser.gemspec`

PR #195, opened 2021-09-10
Fix small and empty HTTP response handling

julik:
If a response is very small (or the HTTP range requested "rolls off" the end
of the resource body) S3 will respond with the full resource and with a 200 status code.
If a resource is empty no Range support will be available (and there is no good way to
address a 0-size resource using the HTTP Ranges specification).

To handle this better we will now allow small 200 responses to be treated as normal responses,
as long as they do not exceed the range size we request from the server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.29.1
+* Fix handling of 200 responses with `parse_http` as well as handling of very small responses which do not need range access
+
 ## 0.29.0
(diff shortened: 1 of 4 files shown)


fabioperrella (approved):

fabioperrella:
@julik I don't know if you noticed but some tests broke for old Rubies

linkyndy (approved):

julik:
Yeah it's the squiggly heredoc. gimme a sec

PR #199, opened 2022-01-10
Use local image for rails test instead of remote

lao:
One of the tests was linked to this image ( https://freesvg.org/img/1416155153.png ), which apparently used to be of dimensions 500x296 and now it is 600x600 that is why some tests are breaking.

Changed to local image, and changed dimensions.


grdw:
How can a change to this image cause a JSON parser error down the line in Ruby 2.3.0? 🤔. That's a bit random, isn't it?

lao:
@grdw that error in ruby 2.3.0 is happening due to some weird behaviour (which I cannot reproduce locally at all), not caused my this PR, but also we were discussing about dropping support to version 2.3.0 until 2.5.0.

PR #201, opened 2022-02-15
Fix undefined method atom_fields for nil exception

enricoribelli:


lib/parsers/moov_parser/decoder.rb
```diff
@@ -72,7 +72,7 @@ def find_video_trak_atom(atoms)
 
     trak_atoms.find do |trak_atom|
       hdlr_atom = find_first_atom_by_path([trak_atom], 'trak', 'mdia', 'hdlr')
```
linkyndy, line 74: This means `hdlr_atom` is `nil`. I haven't checked the format specifications, but would that be a valid scenario? 🤔

lib/parsers/moov_parser/decoder.rb
```diff
@@ -72,7 +72,7 @@ def find_video_trak_atom(atoms)
...
-      hdlr_atom.atom_fields[:component_type] == 'mhlr' && hdlr_atom.atom_fields[:component_subtype] == 'vide'
+      hdlr_atom&.atom_fields[:component_type] == 'mhlr' && hdlr_atom&.atom_fields[:component_subtype] == 'vide'
```
linkyndy, line 75: Looking at the rest of the file, I feel exiting early stays more in line with the coding style adopted. Thus, a `next if hdlr_atom.nil?` would look better.

linkyndy (commented):
I feel it would be worth testing this change as well 😊

PR #202, opened 2022-02-16
Skip hdlr atom when atom is nil

enricoribelli:
Revert latest commit with the atom null check.
Add atom skipping if null.
Add test file with .mov file without hdlr atom.


linkyndy (commented):
Looking good!

As before, it would be nice if you could double check that the format specifications allow such a scenario. And adding a small unit test 😊

lao (approved):

PR #203, opened 2022-03-31
[PREV-37] Add support for frame_rate in moov parser and modify related unit tests

enricoribelli:


lib/parsers/moov_parser.rb
```diff
@@ -115,5 +118,33 @@ def matches_moov_definition?(io)
     maybe_atom_size >= minimum_ftyp_atom_size && maybe_ftyp_atom_signature == 'ftyp'
   end
 
+  def parse_sample_atom(decoder, atom_tree)
+    video_trak_atom = decoder.find_video_trak_atom(atom_tree)
+
+    stts = begin
```
pcriv, line 124: What does `stts` stands for? Also, the `begin` block is unnecessary, you can just do `stts = if video_trak_atom ...`
Same for the code bellow.

lib/parsers/moov_parser.rb
```diff
@@ -48,12 +48,14 @@ def call(io)
...
-    if mdhd = decoder.find_first_atom_by_path(atom_tree, 'moov', 'mvhd')
-      timescale = mdhd.field_value(:tscale)
-      duration = mdhd.field_value(:duration)
+    if mvhd = decoder.find_first_atom_by_path(atom_tree, 'moov', 'mvhd')
+      timescale = mvhd.field_value(:tscale)
+      duration = mvhd.field_value(:duration)
       media_duration_s = duration / timescale.to_f
     end
 
+    frame_rate = parse_sample_atom(decoder, atom_tree).truncate(2)
```
linkyndy, line 57: Would this work for non-video files, at this level?

lib/parsers/moov_parser.rb
```diff
@@ -115,5 +118,33 @@ def matches_moov_definition?(io)
...
+        decoder.find_first_atom_by_path([video_trak_atom], 'trak', 'mdia', 'mdhd')
+      else
+        decoder.find_first_atom_by_path(atom_tree, 'moov', 'trak', 'mdia', 'mdhd')
+      end
+    end
+
+    if stts && mdhd
+      timescale = mdhd.atom_fields[:tscale]
+      sample_duration = stts.field_value(:entries).first[:sample_duration]
+      return timescale.to_f / sample_duration
```
linkyndy, line 143: You could omit the `return` here.

linkyndy, line 143: Are we sure this won't raise any division error? I see a similar operation above, but I wanted to double check.

linkyndy (commented):
Could you please remove the Jira link from the description? It's a public repo 😊

lib/parsers/moov_parser.rb
enricoribelli: those fields must be there if the whole atom is there. However a file can be corrupted or not valid so this is a point too and added extra checks to make sure we are not doing illegal operations

enricoribelli: done

lib/parsers/moov_parser.rb
enricoribelli: it would still work, I tried it with the m4a test file in our fixtures but you still have a point and moved this a little bit further when we know for sure we are dealing with a video.

lib/parsers/moov_parser.rb
enricoribelli: "stts" is the name of an atom in that format, will add a comment to clarify that

lib/parsers/moov_parser.rb
```diff
@@ -115,5 +117,35 @@ def matches_moov_definition?(io)
...
+    mdhd = if video_trak_atom
+      decoder.find_first_atom_by_path([video_trak_atom], 'trak', 'mdia', 'mdhd')
+    else
+      decoder.find_first_atom_by_path(atom_tree, 'moov', 'trak', 'mdia', 'mdhd')
+    end
+
+    if stts && mdhd
+      timescale = mdhd.atom_fields[:tscale]
+      sample_duration = stts.field_value(:entries).first[:sample_duration]
+      if timescale.nil? || timescale == 0 || sample_duration.nil? || sample_duration == 0
```
linkyndy, line 140: You can divide 0 by anything, that shouldn't be an invalid case.

lib/parsers/moov_parser.rb
```diff
@@ -115,5 +117,35 @@ def matches_moov_definition?(io)
...
+        nil
+      else
+        timescale.to_f / sample_duration
```
linkyndy, line 143: How does float division paired with the above truncation work for some odd frame_rates? I suspect we'd get some pretty weird values sometimes. Curious if you tried running some videos through other tools and check what they return for the frame rate.

lib/parsers/moov_parser.rb
enricoribelli: true, but I added that part because diving 0 by anything gives 0, which is not a useful info. In such case I'd rather return a null instead of a "frame_rate = 0"

lib/parsers/moov_parser.rb
enricoribelli: I tried to look for weird videos and I saw in our fixture folder (within the spec folder) has already some videos. One of these has a weird frame rate of 14.98 and the tool calculates that properly.. if this is what you meant!

lib/parsers/moov_parser.rb
linkyndy: Not sure whether the timescale can legitimately be zero, but I trust your judgement on this one 👍

lib/parsers/moov_parser.rb
linkyndy: If you've checked that, then I guess we're fine. It's just odd using `truncate` in the context of floats, was expecting something more in the lines of `round`.

lib/parsers/moov_parser.rb
enricoribelli: legitimately = no
illegitimately = yes, and in that case at least we are covered 😬

lib/parsers/moov_parser.rb
```diff
@@ -64,13 +64,15 @@ def call(io)
         intrinsics: atom_tree,
       )
     else
+      frame_rate = parse_sample_atom(decoder, atom_tree)&.truncate(2)
```
linkyndy, line 67: On second thoughts, do we even need truncating here? I see the media duration is kept as is. Is there a reason we need two digits for this value? 🤔

enricoribelli: wherever I saw a frame rate, it was expressed in the form "ii,dd" . So, whenever I was making the calculation of timescale / sample_duration , the result, if both values are integers is of course an integer, which is not correct in our case. If just one of the 2 becomes a decimal, then we have the correct number but in the format "ii,dddddddd" . By rounding that up we would get the wrong frame rate, while instead truncating it was the thing to do, or at least truncation matched the result given back by the format parser with the frame rate shown by normal video players.

linkyndy: Yeah, meant just leaving the value as is, without truncating or rounding. Duration is also usually given with _some_ precision, however we return the whole thing.

If you feel truncation makes sense in this case, let's go with that.

linkyndy (approved):
Nice effort! 💪

PR #204, opened 2022-04-06
Add support for video codec metadata

enricoribelli:


lib/parsers/moov_parser/decoder.rb
```diff
@@ -130,6 +130,23 @@ def parse_stts_atom(io, _)
...
+    stsd = {
+      version: version,
+      flags: read_bytes(io, 3),
+      number_of_entries: is_v1 ? read_64bit_uint(io) : read_32bit_uint(io),
+      codecs: []
+    }
+    stsd[:number_of_entries].times {
+      codec_length = read_32bit_uint(io)
+      stsd[:codecs] << read_bytes(io, 4)
+      io.seek(io.pos + codec_length - 8) # 8 bytes is the header length containing the codec length and the codec name that we just did read
```
linkyndy, line 145: Why do we have to seek? 🤔

spec/parsers/moov_parser_spec.rb
```diff
@@ -94,6 +94,7 @@ def deep_print_atoms(atoms, output, swimlanes = [])
     expect(result.format).to eq(:mov)
     expect(result.width_px).to eq(1920)
     expect(result.height_px).to eq(1080)
+    expect(result.codecs.first).to eq('apcn')
```
linkyndy, line 97: Shouldn't we check the whole array of codecs we just parsed?

lib/parsers/moov_parser/decoder.rb
enricoribelli: Because the "codec" itself is an atom containing lots of information inside that we don't care about. For example:
image
As you can see in this example **hvc1** is the codec used, and also the info that we need, but beneath it has quite some info we don't care about, so I am skipping all of those to jump to the next codec, if there

linkyndy: Right, now it makes sense. Thanks for the explanation!

spec/parsers/moov_parser_spec.rb
linkyndy: Yes, get that, was just referring to the expectation:

```suggestion
    expect(result.codecs).to eq(['apcn'])
```

Otherwise, it feels the test is only concerned with the first codec, which is not. Matching against the whole array better stands for the point, imho.

enricoribelli: Yes I get your point! Committed the change!

linkyndy: Awesome! 👌

linkyndy (approved):
Great job! 💪

PR #205, opened 2022-04-19
[PREV-51] fix nil sample_duration in MOOV parser

enricoribelli:
fix nil sample_duration in MOOV parser

### DIFF
--- a/lib/parsers/moov_parser.rb
+++ b/lib/parsers/moov_parser.rb
@@ -1,5 +1,3 @@
-require 'pry'
-
 class FormatParser::MOOVParser
   include FormatParser::IOUtils
   require_relative 'moov_parser/decoder'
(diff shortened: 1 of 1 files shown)


lib/parsers/moov_parser.rb
```diff
@@ -136,7 +136,8 @@ def parse_time_to_sample_atom(decoder, atom_tree)
 
     if stts && mdhd
       timescale = mdhd.atom_fields[:tscale]
-      sample_duration = stts.field_value(:entries).first[:sample_duration]
```
linkyndy, line 139: I think you could simplify this by using `dig`: `stts.field_value(:entries).dig(0, :sample_duration)`

enricoribelli: I didn't know about this `dig` thanks!

PR #206, opened 2022-04-20
revolve exception bug happening when the stts atom is nil

enricoribelli:
revolve exception bug happening when the stts atom is not contained in the item during codecs parsing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.2.1
+* Resolve bug when `stts` atom is nil
+
 ## 1.2.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 1.2.1
+* Resolve bug when `stts` atom is nil
```
linkyndy, line 2:
```suggestion
* Resolve bug when `stts` atom is `nil`
```

linkyndy (approved):
Looks good! 💪

PR #207, opened 2022-05-11
Add support HEIF and HEIC format

enricoribelli:
- Add heif_parser
- heif_parser parses both HEIF and HEIC single images and image collections but not image sequences.
- Missing EXIF metadata parsing


lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+  HEADER_LENGTH = 8 # every box header has a length of 8 bytes
+  HEIC_MIME_POSSIBLE_TYPES = {
+    'heic' => :heic,
+    'heix' => :heix,
+    'heim' => :heim,
+    'heis' => :heis
+  }
+  HEIC_MIME_TYPE = 'image/heic'
+  HEIF_MIME_TYPE = 'image/heif'
+  # HEIC_SEQUENCE_MIME_TYPE = 'image/heic-sequence' # TODO: to be used when adding image-sequence parsing
```
pcriv, line 37: Do we need these comments? also, they seem to be duplicated

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+    end
+
+    result = FormatParser::Image.new(
+      format: @format,
+      width_px: @width,
+      height_px: @height,
+      intrinsics: {
+        'compatible_brands': @compatible_brands,
+        'handler_type': @handler_type,
+        # 'sub_items': @sub_items, # enable this if you want to output all the sub-itmes in the image
```
pcriv, line 107: Do we need this too?

pcriv (approved):
Looks like a really complicated feature to add, so kudos @enricoribelli and great job! 💯

lib/parsers/heif_parser.rb
enricoribelli: we don't need them now, no, I left them there to save some time and help the next person who is going to eventually work on this and wants to add support for image sequences.

lib/parsers/heif_parser.rb
enricoribelli: this can be really useful mainly for debugging reasons

lib/parsers/heif_parser.rb
pcriv: Cool but we don't need it twice right?

enricoribelli: they are not the same, one is hei**f** and the other hei**c**

pcriv: oh! my bad

enricoribelli: no problem:)

.gitignore
```diff
@@ -60,4 +60,4 @@ Gemfile.lock
 .DS_Store
 
 # rspec examples
-spec/examples.txt
+spec/examples.txt
```
linkyndy, line 63: New lines at EOF are usually advised 😇

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 1.3.1
+* Add `heif_parser` and support for `HEIF and HEIC` formats. Exif parsing is still missing.
```
linkyndy, line 2:
```suggestion
* Add `heif_parser` and support for `HEIF` and `HEIC` formats. Exif parsing is still missing.
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 1.3.1
```
linkyndy, line 1: I think `1.3.0` makes more sense 😊

lib/parsers/heif_parser.rb
linkyndy, line 37: `# TODO:`s are usually on their own line; at least that's what the Ruby style guide recommends, mostly because code parsers are looking for them there.

lib/parsers/heif_parser.rb
linkyndy, line 107:
```suggestion
        # 'sub_items': @sub_items, # enable this if you want to output all the sub-items in the image
```

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+    )
+
+    result
+  end
+
+  def scan_file_type_box
+    file_type_box_length = read_int_32
+    return unless read_string(4) == FILE_TYPE_BOX_MARKER
+    @major_brand = read_string(4)
+    return unless @major_brand == HEIF_MARKER || MIF1_MARKER
```
linkyndy, line 126: Just wanted to double check whether no parens are needed here ✌️

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+    read_string(4) # minor_brand
+
+    # Subtracting from the total length of the box specified in the header the size header itself (8 bytes = header length and length of ftyp)
+    # and the length of the major and minor brand, we obtain the the compatible brands
```
linkyndy, line 130:
```suggestion
    # and the length of the major and minor brand, we obtain the compatible brands
```

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+      when 3
+        item_id = read_int_32
+      else
+        return # wrong version according to standards, hence return
+      end
+      safe_skip(@buf, 2) # not interested in the item_protection_index
+      item_type = read_string(4)
+      content_encoding = ''
+      if item_type == MIME_MARKER
+        content_encoding = read_string(item_info_end_pos - @buf.pos).delete!("\0") # remove the null-termination part for output visualization reason
```
linkyndy, line 206: Do we need to mutate this string? Often it's at the core of many hard-to-debug issues.

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+      end
+      @sub_items << {'item_id': item_id, 'item_type': item_type, 'content_encoding': content_encoding}
```
linkyndy, line 208: Do you wish these to be symbol keys or string keys? Because you may drop the quotes if you need symbols, or use `=>` if you need strings 😊

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+        height = read_int_32
+        @item_props[item_prop_index] = {
+          'type': IMAGE_SPATIAL_EXTENTS_BOX,
+          'width': width,
+          'height': height
+        }
+      when PIXEL_ASPECT_RATIO_BOX
+        h_spacing = read_int_32
+        v_spacing = read_int_32
+        pixel_aspect_ratio = h_spacing.to_s + '/' + v_spacing.to_s
```
linkyndy, line 262: You could use interpolation to make this less verbose:

```suggestion
        pixel_aspect_ratio = "#{h_spacing}/#{v_spacing}"
```

lib/parsers/heif_parser.rb
```diff
@@ -0,0 +1,430 @@
...
+    end
+    binary_value = binary.reverse
+    (8 - binary_value.length).times do
+      binary_value.prepend('0')
+    end
+    binary_value
+  end
+
+  def likely_match?(filename)
+    filename =~ /\.(heif|heic)$/i
```
linkyndy, line 426: You can shorten this a bit:

```suggestion
    filename =~ /\.hei[cf]$/i
```

lib/parsers/moov_parser.rb
```diff
@@ -163,5 +163,5 @@ def parse_sample_description_atom(decoder, atom_tree)
     end
   end
 
-  FormatParser.register_parser new, natures: :video, formats: FTYP_MAP.values, priority: 1
```
linkyndy, line 166: Curious why this needs to be changed; I thought parsers should be listed in the order of likelihood of a random file matching it 🤔

linkyndy (commented):
Great effort, Enrico! 💪   Left a few small comments/questions.

lib/parsers/heif_parser.rb
enricoribelli: the content is the same, I just remove some trailing zeros at the end but the real content is left untouched

lib/parsers/heif_parser.rb
enricoribelli: good tip! Thank you! :)

lib/parsers/moov_parser.rb
enricoribelli: @linkyndy  there are currently 2 ways of matching a file to a certain parser, either you pass the format itself (e.g. "heic") in the format_parser as input alongside the file itself and then format_parser thanks to the **likely_match** method knows which parser to use, or you leave it up to the format_parser itself but priorities take place then. A HEIC file, due to its nature, could be associated (wrongly) to a .moov file and therefore if priorities take place then the moov parser would be used, which we don't want to. So I wanted to make space for the heif_parser to be in terms of priorities before the moov_parser, by at the same time making sure that nothing is broken on the moov_parser side.

lib/parsers/heif_parser.rb
enricoribelli: @linkyndy  nope but I just noticed that it gives a warning so I changed it to `return unless @major_brand == HEIF_MARKER || @major_brand == MIF1_MARKER`

linkyndy: I think that's a bit clearer 👍  You could also do `[FOO, BAR].includes?(baz)`, if you find that more readable.

lib/parsers/heif_parser.rb
linkyndy: Sure, was just wondering whether a simple `String#delete` wouldn't do the trick.

lib/parsers/moov_parser.rb
linkyndy: Ah, interesting! Thanks for explaining.

lib/parsers/heif_parser.rb
enricoribelli: meeh, less code, true, but I think less immediate.. would leave it like this if you don't mind!

linkyndy: Sure!

linkyndy (approved):
Great job! 🚀

PR #209, opened 2022-06-29
WebP File Format Support

### LINKED ISSUE #57: WEBP format support
Dimensions, color mode and whether it has transparency

Kevin-McGonigle:
* Closes #57.

* Adds a parser (and accompanying tests) for parsing WebP lossy, lossless and extended files, obtaining the following info:
  * Height
  * Width
  * Transparency?
  * Animated?
  * Exif Metadata
  * XMP Metadata
  * Number of frames (if animated)


enricoribelli:
With regards to the fixtures you are using for the test, where did you get them from? Just making sure they are free of copyright and therefore safe to be used on our side for testing!

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,123 @@
+# WebP is an image format that provides superior lossless and lossy compression for images on the web, with support for
+# transparency. It uses predictive coding to encode an image, predicting the values in a block of pixels based on the
+# values of neighbouring blocks. A WebP file consists of VP8 or VP8L data, and a container based on RIFF. There is also
+# an extended file format, VP8X, that optionally encodes various information such as the color profile, animation
+# control data, transparency, and EXIF and/or XMP metadata.
+#
+# For more information, visit https://developers.google.com/speed/webp.
+#
+# TODO: Extract Exif and XMP metadata.
```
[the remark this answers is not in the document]
enricoribelli: Ps: if at the end there is work left for not releasing the full parsing of the format let's just remember to create the related issue on the repo itself, so that who wants can pick it up

Kevin-McGonigle:
> With regards to the fixtures you are using for the test, where did you get them from? Just making sure they are free of copyright and therefore safe to be used on our side for testing!

@enricoribelli I left a little note about that in the [README](https://github.com/WeTransfer/format_parser/pull/209/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R202-R205). The animated one is from Wiki Commons with a Creative Commons licence, similar to other fixtures elsewhere. The remaining fixtures are just pictures of my cat that I've edited to fit the test cases 😂

spec/parsers/webp_parser_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  it 'successfully parses extended WebP files with animation' do
+    result = subject.call(File.open(fixtures_dir + 'WEBP/extended-animation.webp', 'rb'))
+    check_result(result, true, 211, 211, true, true)
+  end
+
+  private
+
+  def check_result(
+    result,
```
enricoribelli, line 47: not sure about DRYing up the code this much. I understand that most of the things being checked by these tests are fairly the same, thus you are trying to avoid repetition. But the moment for example you start adding extra stuff, like an EXIF, and you'll want to test it, you'll need to add an extra argument in here as "expected behaviour".. and then one more and one more again, risking to make the method declaration quite long.

Plus it requires you going back and forward to see "oh what did I pass to this method? (scroll up) Ah yes I passed TRUE (scroll down). And what was the height_px? (scroll up).. etc" . I'd rather have a test well compacted in one place and that's it. Helpers method re-used can be extracted of course and declared separately, but this is not one of those. This is actual logic being checked.

enricoribelli:
whoop whoop your first PR ! 🥳

.gitignore
```diff
@@ -61,3 +61,6 @@ Gemfile.lock
 
 # rspec examples
 spec/examples.txt
+
+# IntelliJ config:
+/.idea/
```
linkyndy, line 66: Not sure about this one; but at least add a blank line at EOF 😇

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
+## 1.4.0
+* Add support for WebP lossy, lossless and extended file formats.
```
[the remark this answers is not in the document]
linkyndy: Would say `WEBP` even, given the other references in this PR 😊

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,123 @@
...
+    #   - I = Set if file contains an ICC profile.
+    #   - L = Set if file contains transparency information.
+    #   - E = Set if file contains Exif metadata.
+    #   - X = Set if file contains XMP metadata.
+    #   - A = Set if file is an animated image.
+    flags = safe_read(@io, 1).unpack('C')[0]
+    has_transparency = flags & 0x10 != 0
+    has_multiple_frames = flags & 0x02 != 0
+
+    # Note: num_animation_or_video_frames cannot be calculated without parsing the entire file.
```
linkyndy, line 99:
```suggestion
    # NOTE: num_animation_or_video_frames cannot be calculated without parsing the entire file.
```

spec/parsers/webp_parser_spec.rb
linkyndy: What Enrico said ☝️ I support the idea of being more declarative and less _smart_ when it comes to writing tests, as it improves readability, which is arguably the most important aspect with tests.

linkyndy (commented):
Few comments, but your (first!) work looks good! 🙌

spec/parsers/webp_parser_spec.rb
Kevin-McGonigle: Fair point. I do have a tendency to lean towards DRYness in situations like this, probably a holdover from working in payments code where there are lots of lovely little edge cases where almost everything else remains the same 😅. It'll hurt my heart a bit, but good shout, I'll switch it up 👍🏻 .

lib/parsers/webp_parser.rb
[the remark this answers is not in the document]
enricoribelli: I would say give it a try and let's see where you get at. It would be nice to release full parser whenever possible! I understand it can be very cumbersome, I had this experience myself not long ago, but so far luckily this format doesn't look too complex!
What do you mean the whole file needs to be parsed? If for example this kind of format is the type of "content header says the next 600bytes are content and then we find the EXIF/XMP" then I'd definitely go for it.
If instead you have to go through 20 different internal headers in the content, and scrutinize each one of them for a chance of a nested EXIF, then it's another story. Let me know!

lib/parsers/webp_parser.rb
```diff
@@ -119,5 +126,35 @@ def create_image(width, height, has_multiple_frames: false, has_transparency: fa
...
+        exif ||= exif_from_tiff_io(StringIO.new(safe_read(@buf, chunk_size)))
+      when 'XMP'
+        xmp ||= parse_xmp_metadata(chunk_size)
+      when 'ANMF'
+        num_frames += 1 if has_multiple_frames
+        safe_skip(@buf, chunk_size)
+      else
+        safe_skip(@buf, chunk_size)
+      end
+    end rescue FormatParser::IOUtils::InvalidRead
```
enricoribelli, line 148: could you move the `rescue FormatParser::IOUtils::InvalidRead` on a new line? we blind developers like good indentation or we tend to miss stuff 🤓

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,160 @@
...
+    image.height_px, image.width_px = image.width_px, image.height_px if exif&.rotated?
+    if exif || xmp
+      image.intrinsics = {}
+      image.intrinsics[:exif] = exif if exif
+      image.intrinsics[:xmp] = xmp if xmp
+    end
+    image.orientation = exif&.orientation_sym
```
enricoribelli, line 155: why not placing this in the **if** above since there we are checking for `nil` anyway already?

lib/parsers/webp_parser.rb
```diff
@@ -119,5 +126,35 @@ def create_image(width, height, has_multiple_frames: false, has_transparency: fa
...
```
enricoribelli, line 155: why not placing this in the`if` above since there we are checking for `nil` anyway already?

lib/parsers/webp_parser.rb
```diff
@@ -119,5 +126,35 @@ def create_image(width, height, has_multiple_frames: false, has_transparency: fa
     )
   end
 
+  def augment_image(image)
+    # We're going to scan the file looking for the EXIF, XMP and/or ANMF chunks.
+    exif = nil
+    xmp = nil
+    num_frames = 0
+    loop do
```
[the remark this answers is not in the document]
Kevin-McGonigle: Yeah I'm doing that already - if you look inside the case statement, I use `safe_skip`/`safe_read` to skip over chunks that aren't relevant and parse chunks that are relevant, respectively.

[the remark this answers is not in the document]
enricoribelli: missed it sorry! perfect :)

lib/parsers/webp_parser.rb
```diff
@@ -119,5 +126,35 @@ def create_image(width, height, has_multiple_frames: false, has_transparency: fa
...
+      chunk_code, chunk_size = safe_read(@buf, 8).unpack('A4V')
```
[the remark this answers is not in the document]
Kevin-McGonigle: Had to go with `fourcc` to appease Rubocop.

lib/parsers/webp_parser.rb
enricoribelli: that's correct! It wants to indent it there because it's where it should be :)
when you do:
```
def my_method
  do_something()
rescue ..
  do_something_else()
end
```

it's like you are doing:

```
def my_method
  begin
    do_something()
  rescue
    do_something_else()
  end
end
```

(we don't want the 2nd one because we want to use implicit `begin`s).

lib/parsers/webp_parser.rb
```diff
@@ -132,27 +127,30 @@ def augment_image(image)
...
     if exif || xmp
       image.intrinsics = {}
-      image.intrinsics[:exif] = exif if exif
+      if exif
+        image.height_px, image.width_px = image.width_px, image.height_px if exif.rotated?
+        image.intrinsics[:exif] = exif
+        image.orientation = exif.orientation_sym
+      end
       image.intrinsics[:xmp] = xmp if xmp
     end
```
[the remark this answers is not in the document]
Kevin-McGonigle: If we're cool with `intrinsics` being an empty hash instead of `nil`, then it becomes trivial:
```
image.intrinsics = {}
if exif
  image.height_px, image.width_px = image.width_px, image.height_px if exif.rotated?
  image.intrinsics[:exif] = exif
  image.orientation = exif.orientation_sym
end
image.intrinsics[:xmp] = xmp if xmp
```
but that may well not be the case.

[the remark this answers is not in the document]
enricoribelli: thought EXIF and XMP where mutually exclusive! But if you say they can be both be there then sure, definitely fine with it! Only thing about the last option: check out the other formats, just to make sure to be consistent. If other formats do also empty hash, go for it. If instead they all do `nil` then let's keep it consistent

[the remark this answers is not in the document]
Kevin-McGonigle: Yeah I thought they might be too, but after checking some of my own photos, they're definitely both there! Awesome, I'll check. Thanks!

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  def augment_image(image)
+    # We're going to scan the file looking for the EXIF, XMP and/or ANMF chunks.
+    exif = nil
+    xmp = nil
+    num_frames = 0
+    loop do
+      fourcc, chunk_size = safe_read(@buf, 8).unpack('A4V')
+      safe_skip(@buf, 1) if chunk_size.odd? # Padding byte of 0 added if chunk size is odd.
```
linkyndy, line 133: I'd keep comments on their own line, for readability purposes. I'd also add some new lines in-between logical bits of code, for the same reason.

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,159 @@
...
+      case fourcc
+      when 'EXIF'
+        exif ||= exif_from_tiff_io(StringIO.new(safe_read(@buf, chunk_size)))
```
linkyndy, line 136: Why use `||=`? 🤔

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,159 @@
...
+      when 'ANMF'
+        num_frames += 1 if image.has_multiple_frames
+        safe_skip(@buf, chunk_size)
+      else
+        safe_skip(@buf, chunk_size)
+      end
+    end
+  rescue FormatParser::IOUtils::InvalidRead
+    intrinsics = {}
+    if exif
```
linkyndy, line 148: Looking at the JPG parser, it seems that EXIF data is always set in the intrinsics, even when it's empty. Maybe we can do the same, for consistency reasons.

format_parser.gemspec
```diff
@@ -36,6 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'faraday', '~> 0.13'
   spec.add_dependency 'faraday_middleware', '~> 0.14'
   spec.add_dependency 'measurometer', '~> 1'
+  spec.add_dependency 'nokogiri', '~> 1'
```
linkyndy, line 39: Oh noes, please don't add nokogiri 🙈  Do we _actually_ need it?

lib/parsers/webp_parser.rb
Kevin-McGonigle: According to the [WebP spec](https://developers.google.com/speed/webp/docs/riff_container#metadata), there "should" be at most one of each Exif/XMP chunk, and any such chunks after the first "may" be ignored. That "may" is, of course, ambiguous and leaves it up to us, so we can choose to interpret it differently if you think it's better to.

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,159 @@
...
+      width_px: width
+    )
+  end
+
+  def augment_image(image)
+    # We're going to scan the file looking for the EXIF, XMP and/or ANMF chunks.
+    exif = nil
+    xmp = nil
+    num_frames = 0
+    loop do
```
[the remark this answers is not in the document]
Kevin-McGonigle: Yeah I would've liked to have done that, but `safe_read` will throw that exception when it reaches EOF. I could explicitly call `@buf.read(8)` instead and break if that comes back as `nil` instead?

format_parser.gemspec
Kevin-McGonigle: That comes back to my question about what we want to do about XMP 🤷🏻 If we want XMP to be parsed into accessible object as we do with Exif, then I think we'll need something like nokogiri. We could store the XMP as a string though and defer the responsibility of parsing it to the caller?

lib/parsers/webp_parser.rb
linkyndy: Right 👍  Then maybe a small comment would add more clarity to why we're doing this.

lib/parsers/webp_parser.rb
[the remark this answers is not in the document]
linkyndy: That would be okay. Or wrap the error handling around `safe_read` and break the loop in the rescue block (so you reuse `safe_read`).

lib/parsers/webp_parser.rb
```diff
@@ -0,0 +1,162 @@
...
+
+  def augment_image(image)
+    # We're going to scan the file looking for the EXIF, XMP and/or ANMF chunks.
+    intrinsics = {}
+    num_frames = 0
+    loop do
+      # Try to read the next chunk header, and break the loop if we've reached EOF.
+      begin
+        fourcc, chunk_size = safe_read(@buf, 8).unpack('A4V')
+      rescue InvalidRead
```
linkyndy, line 132: Just wanted to double check if this constant is properly resolved (without providing its full scope).

spec/parsers/webp_parser_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  it 'successfully parses extended WebP files with Exif metadata' do
+    result = subject.call(File.open(fixtures_dir + 'WEBP/extended-exif.webp', 'rb'))
+    expect(result).not_to be_nil
+    expect(result.content_type).to eq('image/webp')
+    expect(result.format).to eq(:webp)
+    expect(result.has_multiple_frames).to eq(false)
+    expect(result.has_transparency).to eq(false)
+    expect(result.height_px).to eq(181)
+    expect(result.intrinsics).not_to be_nil
```
linkyndy, line 87: You could do this in a more expressive way with:

```
expect(...).to match(exif: { image_length: '...', ... })
```

This would ensure a hash-like match on the intrinsics, which would merge your 4 expectations into one.

linkyndy (approved):
Looking great! 🚢

lib/parsers/webp_parser.rb
Kevin-McGonigle: The tests all pass, so apparently so. Are you suggesting that I move the rest of the loop's body inside this block?

spec/parsers/webp_parser_spec.rb
Kevin-McGonigle: Ooh that's cool! It seems like doing it this way would require matching every single Exif property though (there are 39 of them in this image 😅)

linkyndy: Ah, in that case you could slip in a `a_hash_including()` and match like that 😊

PR #210, opened 2022-07-12
WebP - Handle MalformedTIFF Error.

Kevin-McGonigle:
We are seeing some errors that look like: `EXIFR::MalformedTIFF, no byte order information found`. In the event of these errors, we want to skip the Exif chunk that we are attempting to parse (consistent with CR2, JPEG and TIFF parsers).


linkyndy (approved):
🚀

PR #211, opened 2022-07-12
Prevent MP3Parser parsing WebP files.

Kevin-McGonigle:
We have an edge case where the MP3Parser is running and returning something _"meaningful"_ for WebP files, causing them to be mislabled as MP3 files and incorrectly parsed. This PR adds protection against that edge case, similar to the protections added for similar cases with PNG and ZIP files.

### DIFF
--- a/lib/parsers/mp3_parser.rb
+++ b/lib/parsers/mp3_parser.rb
@@ -28,6 +28,7 @@ class FormatParser::MP3Parser
   # For some edge cases
   ZIP_LOCAL_ENTRY_SIGNATURE = "PK\x03\x04\x14\x00".b
   PNG_HEADER_BYTES = [137, 80, 78, 71, 13, 10, 26, 10].pack('C*')
+  WEBP_HEADER_REGEX = /RIFF.{4}WEBP/i
 
   MAGIC_LE = [0x49, 0x49, 0x2A, 0x0].pack('C4')
   MAGIC_BE = [0x4D, 0x4D, 0x0, 0x2A].pack('C4')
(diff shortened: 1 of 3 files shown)


lib/parsers/mp3_parser.rb
```diff
@@ -68,9 +69,10 @@ def call(raw_io)
     # will terminate here. Same with PNGs. In the future
     # we should implement "confidence" for MP3 as of all our formats
     # it is by far the most lax.
-    header = safe_read(io, 8)
+    header = safe_read(io, 12)
     return if header.start_with?(ZIP_LOCAL_ENTRY_SIGNATURE)
     return if header.start_with?(PNG_HEADER_BYTES)
+    return if header.start_with?(WEBP_HEADER_REGEX)
```
linkyndy, line 75: Looks great! My only concern is that regexes are only [introduced](https://ruby-doc.org/core-2.5.1/String.html#method-i-start_with-3F) with this method in Ruby 2.6. I see we still support Ruby 2.5. Not sure how the tests passed, but could you double check this locally, with Ruby 2.5, with a file that satisfies this condition?

Kevin-McGonigle: It looks to me from that link you sent as if there was a case for regexes pre-2.6:
```
case T_REGEXP:
    {
        bool r = rb_reg_start_with_p(tmp, str);
        if (r) return Qtrue;
    }
```

Installing 2.5 now to be sure though!

Kevin-McGonigle: Confirmed that it works with 2.5.

linkyndy: Appreciate you checking it 🙏

linkyndy (approved):

enricoribelli:
Now it does return the correct metadata when you run it locally (not through unit test) as we discussed in chat, right?

enricoribelli (approved):

Issue #34, opened 2017-12-27
NEF Parser


bermannoah:
NEF files are basically TIFFs with some additional content in the headers. They're a bit of a pain to parse, because libraries like ImageMagick will often confuse them for TIFFs and start throwing errors. It'd be great to have a (relatively) surefire way of determining if a TIFF is actually an NEF, and then parsing the same info we parse for pure TIFFS - resolution and orientation. It'd also be interesting to see if there's something we can do with the preview image usually embedded in the file. 

[Here](http://lclevy.free.fr/nef/) are some docs for the format, and you can find some creative-commons licensed NEF files [here](http://www.rawsamples.ch/index.php/en/nikon).

PR #213, opened 2022-08-01
Adding support for nikon NEF files

rcpalacio:
This PR adds a new parser instead of extending TIFF and adding more rules to it - similarly to what we have for CR2.
There is also some extra complexity added to EXIFParser - which now needs to be able to dissect subIDFs within TIFF files.

This PR closes [Issue 34](https://github.com/WeTransfer/format_parser/issues/34)

I do have a couple questions 

- About fixture sizes: Is it okay to have large fixtures like the ones I've added here? the reason for this is: details from NEF files can vary slightly from camera to camera and configuration. Using an array of samples like this allows us to catch edge cases that we would have left uncovered otherwise.

- About IO budget: NEF parser needs a few more seek operations then TIFF, for instance, in order to fetch the correct dimensions from an image. As an example, that results in reading around 59k from the 70Mb fixture we have. Is this figure acceptable? For reference, TIFF parser reads around 9k from a 10Mb file.

### DIFF
--- a/lib/parsers/exif_parser.rb
+++ b/lib/parsers/exif_parser.rb
@@ -1,5 +1,5 @@
-require "exifr/tiff"
-require "delegate"
+require 'exifr/tiff'
+require 'delegate'
 
 module FormatParser::EXIFParser
   ORIENTATIONS = {
@@ -11,7 +11,7 @@ module FormatParser::EXIFParser
     5 => :left_top,
     6 => :right_top,
     7 => :right_bottom,
-    8 => :left_bottom,
+    8 => :left_bottom
   }
 
   # EXIFR kindly requests the presence of a few more methods than what our IOConstraint
@@ -19,7 +19,7 @@ module FormatParser::EXIFParser
   class IOExt < SimpleDelegator
     def readbyte
       if byte = read(1)
-        byte.unpack("C").first
+        byte.unpack('C').first
       end
     end
 
@@ -175,7 +175,6 @@ module FormatParser::EXIFParser
   EXIFR.logger = Logger.new(nil)
 
   def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
-    # TODO: qq. where should we be rescuing unexpected errors? here or somewhere higher up?
     Measurometer.instrument("format_parser.EXIFParser.exif_from_tiff_io") do
       extended_io = IOExt.new(constrained_io)
       raw_exif_data = EXIFR::TIFF.new(extended_io)
--- a/lib/parsers/exif_parser.rb
+++ b/lib/parsers/exif_parser.rb
@@ -1,5 +1,5 @@
-require 'exifr/tiff'
-require 'delegate'
+require "exifr/tiff"
+require "delegate"
 
 module FormatParser::EXIFParser
   ORIENTATIONS = {
@@ -11,7 +11,7 @@ module FormatParser::EXIFParser
     5 => :left_top,
     6 => :right_top,
     7 => :right_bottom,
-    8 => :left_bottom
+    8 => :left_bottom,
   }
 
   # EXIFR kindly requests the presence of a few more methods than what our IOConstraint
@@ -19,7 +19,7 @@ module FormatParser::EXIFParser
   class IOExt < SimpleDelegator
     def readbyte
       if byte = read(1)
-        byte.unpack('C').first
+        byte.unpack("C").first
       end
     end
 
@@ -41,6 +41,13 @@ module FormatParser::EXIFParser
   end
 
   class EXIFResult < SimpleDelegator
+    attr_reader :sub_ifds_data
+
+    def initialize(exif_raw_data, sub_ifds_data = {})
+      super(exif_raw_data)
+      @sub_ifds_data = sub_ifds_data
+    end
+
     def rotated?
       orientation.to_i > 4
     end
@@ -167,10 +174,37 @@ module FormatParser::EXIFParser
   # Squash exifr's invalid date warning since we do not use that data.
   EXIFR.logger = Logger.new(nil)
 
-  def exif_from_tiff_io(constrained_io)
-    Measurometer.instrument('format_parser.EXIFParser.exif_from_tiff_io') do
-      raw_exif_data = EXIFR::TIFF.new(IOExt.new(constrained_io))
-      raw_exif_data ? EXIFResult.new(raw_exif_data) : nil
+  def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
+    # TODO: qq. where should we be rescuing unexpected errors? here or somewhere higher up?
+    Measurometer.instrument("format_parser.EXIFParser.exif_from_tiff_io") do
+      extended_io = IOExt.new(constrained_io)
+      raw_exif_data = EXIFR::TIFF.new(extended_io)
+      sub_ifds_hash = {}
+
+      if should_include_sub_ifds
+        sub_ifds_offsets = raw_exif_data.map(&:sub_ifds).flatten.compact
+        sub_ifds_hash = load_sub_ifds(extended_io, sub_ifds_offsets)
+      end
+
+      return raw_exif_data ? EXIFResult.new(raw_exif_data, sub_ifds_hash) : nil
+    end
+  end
+
+  private
+
+  # Reads exif data from subIFDs. This is important for NEF files.
+  def load_sub_ifds(extended_io, sub_ifds_offsets)
+    # Returning an hash of subIFDs using offsets as keys
+    # {
+    #    123 => { subIFD data...}
+    #    456 => { another subIFD data...}
+    # }
+    unless sub_ifds_offsets.empty?
+      EXIFR::TIFF::Data.open(extended_io) do |data|
+        return sub_ifds_offsets.map do |sub_ifd_offset|
+                 [sub_ifd_offset, EXIFR::TIFF::IFD.new(data, sub_ifd_offset)]
+               end.to_h
+      end
     end
   end
 
--- a/lib/parsers/exif_parser.rb
+++ b/lib/parsers/exif_parser.rb
@@ -175,7 +175,7 @@ module FormatParser::EXIFParser
   EXIFR.logger = Logger.new(nil)
 
   def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
-    Measurometer.instrument("format_parser.EXIFParser.exif_from_tiff_io") do
+    Measurometer.instrument('format_parser.EXIFParser.exif_from_tiff_io') do
       extended_io = IOExt.new(constrained_io)
       raw_exif_data = EXIFR::TIFF.new(extended_io)
       sub_ifds_hash = {}
--- a/lib/parsers/exif_parser.rb
+++ b/lib/parsers/exif_parser.rb
@@ -201,9 +201,9 @@ module FormatParser::EXIFParser
     return {} if sub_ifds_offsets.empty?
 
     EXIFR::TIFF::Data.open(extended_io) do |data|
-        return sub_ifds_offsets.map do |sub_ifd_offset|
-                [sub_ifd_offset, EXIFR::TIFF::IFD.new(data, sub_ifd_offset)]
-              end.to_h
+      sub_ifds_offsets.map do |sub_ifd_offset|
+        [sub_ifd_offset, EXIFR::TIFF::IFD.new(data, sub_ifd_offset)]
+      end.to_h
     end
   end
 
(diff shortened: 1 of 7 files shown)


lib/parsers/exif_parser.rb
```diff
@@ -167,10 +174,36 @@ def method_missing(*a)
...
+      extended_io = IOExt.new(constrained_io)
+      raw_exif_data = EXIFR::TIFF.new(extended_io)
+      sub_ifds_hash = {}
+
+      if should_include_sub_ifds
+        sub_ifds_offsets = raw_exif_data.map(&:sub_ifds).flatten.compact
+        sub_ifds_hash = load_sub_ifds(extended_io, sub_ifds_offsets)
+      end
+
+      return raw_exif_data ? EXIFResult.new(raw_exif_data, sub_ifds_hash) : nil
```
linkyndy, line 188: You don't necessarily need to specify `return`, as the last line of a method/block will be returned 😊

lib/parsers/exif_parser.rb
```diff
@@ -167,10 +174,36 @@ def method_missing(*a)
...
+  def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
     Measurometer.instrument('format_parser.EXIFParser.exif_from_tiff_io') do
-      raw_exif_data = EXIFR::TIFF.new(IOExt.new(constrained_io))
-      raw_exif_data ? EXIFResult.new(raw_exif_data) : nil
+      extended_io = IOExt.new(constrained_io)
+      raw_exif_data = EXIFR::TIFF.new(extended_io)
+      sub_ifds_hash = {}
+
+      if should_include_sub_ifds
+        sub_ifds_offsets = raw_exif_data.map(&:sub_ifds).flatten.compact
```
linkyndy, line 184: You can also check `flat_map` here.

lib/parsers/exif_parser.rb
```diff
@@ -167,10 +174,36 @@ def method_missing(*a)
...
+  private
+
+  # Reads exif data from subIFDs. This is important for NEF files.
+  def load_sub_ifds(extended_io, sub_ifds_offsets)
+    # Returning an hash of subIFDs using offsets as keys
+    # {
+    #    123 => { subIFD data...}
+    #    456 => { another subIFD data...}
+    # }
+    unless sub_ifds_offsets.empty?
```
linkyndy, line 201: To get rid of a level of indentation, you can just return early: `return if sub_ifds_offsets.empty?`.

lib/parsers/exif_parser.rb
```diff
@@ -167,10 +174,36 @@ def method_missing(*a)
...
+
+      return raw_exif_data ? EXIFResult.new(raw_exif_data, sub_ifds_hash) : nil
+    end
+  end
+
+  private
+
+  # Reads exif data from subIFDs. This is important for NEF files.
+  def load_sub_ifds(extended_io, sub_ifds_offsets)
+    # Returning an hash of subIFDs using offsets as keys
```
linkyndy, line 196: Is `nil` a valid return value for this method?

linkyndy (commented):
Regarding your questions:

> About fixture sizes: Is it okay to have large fixtures like the ones I've added here? the reason for this is: details from NEF files can vary slightly from camera to camera and configuration. Using an array of samples like this allows us to catch edge cases that we would have left uncovered otherwise.

I don't think fixture size should be a reason not to add fixtures/test specific features. The aim is to have an as-complete-as-possible-test-suite, and fixtures should help with this.

> About IO budget: NEF parser needs a few more seek operations then TIFF, for instance, in order to fetch the correct dimensions from an image. As an example, that results in reading around 59k from the 70Mb fixture we have. Is this figure acceptable? For reference, TIFF parser reads around 9k from a 10Mb file.

Looking at the numbers, they are quite proportional, right? If that needs to happen, it should be fine. Would this happen to all TIFs, even if they are not NEFs?

Linked to that, curious about the ordering of the parsers. We've seen in the past parsers recognising a different format because of this ordering. Would we cover everything as expected, from most granular to the least?

rcpalacio:
@linkyndy,
> Looking at the numbers, they are quite proportional, right? If that needs to happen, it should be fine. Would this happen to all TIFs, even if they are not NEFs?

No, the behaviour should be unchanged for regular TIFF files. `exif_from_tiff_io` now accepts an optional parameter `should_include_sub_ifds` to dive into subIDFs and it defaults to **false**.
 
> Linked to that, curious about the ordering of the parsers. We've seen in the past parsers recognising a different format because of this ordering. Would we cover everything as expected, from most granular to the least?

That is a good catch! Ordering is fine because NEF comes before TIFF alphabetically, but to rely on that is quite weak indeed. I'll add `priority: 4` to NEF's `register_parser` to make it explicit 😄

lib/parsers/exif_parser.rb
rcpalacio: good one 👍

lib/parsers/exif_parser.rb
rcpalacio: It was, yes, but it's still a fair point. I changed it now to return an empty hash instead and make subIFD data in `EXIFResult` a bit more consistent and predictable.

lib/parsers/exif_parser.rb
```diff
@@ -198,12 +198,12 @@ def load_sub_ifds(extended_io, sub_ifds_offsets)
     #    123 => { subIFD data...}
     #    456 => { another subIFD data...}
     # }
-    unless sub_ifds_offsets.empty?
-      EXIFR::TIFF::Data.open(extended_io) do |data|
+    return {} if sub_ifds_offsets.empty?
+
+    EXIFR::TIFF::Data.open(extended_io) do |data|
```
linkyndy, line 203: Indentation is a bit weird here 😇  Also, you probably don't need `return` here either.

spec/parsers/exif_parser_spec.rb
```diff
@@ -1,15 +1,88 @@
...
+
+        path = fixtures_dir + 'NEF/RAW_NIKON_D40_SRGB.NEF'
+        should_include_sub_ifds = true
+
+        exif_data = File.open(path, 'rb') do |f|
+          described_class.exif_from_tiff_io(f, should_include_sub_ifds)
+        end
+
+        offset_1, offset_2 = exif_data.sub_ifds.sort
+        first_sub_ifd = exif_data.sub_ifds_data&.[](offset_1)
```
linkyndy, line 79: If we have the actual file and thus know its EXIF data, why do we still have to guard ourselves with `&.`?

spec/parsers/nef_parser_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+        expect(parsed.intrinsics[:exif]).not_to be_nil
+      end
+    end
+  end
+
+  describe 'Image Dimensions' do
+    it 'parses dimensions properly for a given fixture' do
+      # The default parser from EXIFr returns the dimensions from the embedded
+      # thumbnails as being the image's actual dimensions.
+      # We make sure we properly deal this.
```
linkyndy, line 30:
```suggestion
      # We make sure we properly deal with this.
```

spec/parsers/exif_parser_spec.rb
rcpalacio: This tries to make makes sure that, if we mess and break `exif_from_tiff_io` in the future, the errors will be raised by the assertions and not by this line -- making it easier to follow at first glance and better to maintain.

lib/parsers/nef_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def call(io)
+    io = FormatParser::IOConstraint.new(io)
+
+    return unless HEADER_BYTES.include?(safe_read(io, 4))
+
+    # Because of how NEF files organize their IFDs and subIFDs, we need to dive into the subIFDs
+    # to get the actual image dimensions instead of the preview's
+    should_parse_sub_ifds = true
```
enricoribelli, line 24: I'd say this is a constant because it's always gonna be true for the nef_parser. So I'd put it amongst the other constants on top. What do you think? 🤔

lib/parsers/nef_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+  def valid?(exif_data)
+    # NEF files should hold subIFDs and have "NIKON" or "NIKON CORPORATION" as maker
+    has_sub_ifds_data = !exif_data&.sub_ifds_data&.keys.empty?
+    has_sub_ifds_data && exif_data&.make&.start_with?('NIKON')
+  end
+
+  # Investigates data from all subIFDs and find the one holding the full-resolution image
+  def get_full_resolution_ifd(exif_data)
+    # Most of the time, NEF files have 2 subIFDs:
+    # First one: Thumbnail (Reduce resolution)
```
enricoribelli, line 58: 'Reduced' 😬  ..sorry... I know..

lib/parsers/exif_parser.rb
```diff
@@ -41,6 +41,13 @@ def seek(n, seek_mode = IO::SEEK_SET)
   end
 
   class EXIFResult < SimpleDelegator
+    attr_reader :sub_ifds_data
+
+    def initialize(exif_raw_data, sub_ifds_data = {})
```
enricoribelli, line 46: could we by any chance standardize the naming? you pass them in as `raw_exif_data, sub_ifds_hash` but they become `exif_raw_data, sub_ifds_data`. Would be nice to have them everywhere referenced in the same way 😊

lib/parsers/nef_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+      intrinsics: { exif: exif_data },
+      content_type: NEF_MIME_TYPE,
+    )
+  rescue EXIFR::MalformedTIFF
+    nil
+  end
+
+  def valid?(exif_data)
+    # NEF files should hold subIFDs and have "NIKON" or "NIKON CORPORATION" as maker
+    has_sub_ifds_data = !exif_data&.sub_ifds_data&.keys.empty?
```
enricoribelli, line 51: there is no chance at all for what I see that `sub_ifds_data` is `nil`,  right? Can be empty, but not `nil`? 🤔

lib/parsers/nef_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+    has_sub_ifds_data && exif_data&.make&.start_with?('NIKON')
```
enricoribelli, line 52: we already checked if `exif_data` is `nil` in the line right above.. so we don't need `&` here

lib/parsers/nef_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    return unless HEADER_BYTES.include?(safe_read(io, 4))
+
+    # Because of how NEF files organize their IFDs and subIFDs, we need to dive into the subIFDs
+    # to get the actual image dimensions instead of the preview's
+    should_parse_sub_ifds = true
+
+    exif_data = exif_from_tiff_io(io, should_parse_sub_ifds)
+
+    return unless valid?(exif_data)
```
enricoribelli, line 28: do we have unit tests that check the result when the data is not valid? I may have overlooked 🙏

spec/parsers/exif_parser_spec.rb
linkyndy: I see your point, but on those grounds, anything could break in the future and I'd argue broken code should surface from expectations and not just...break 😊

lib/parsers/exif_parser.rb
```diff
@@ -167,10 +174,36 @@ def method_missing(*a)
...
+  def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
     Measurometer.instrument('format_parser.EXIFParser.exif_from_tiff_io') do
-      raw_exif_data = EXIFR::TIFF.new(IOExt.new(constrained_io))
-      raw_exif_data ? EXIFResult.new(raw_exif_data) : nil
+      extended_io = IOExt.new(constrained_io)
+      raw_exif_data = EXIFR::TIFF.new(extended_io)
+      sub_ifds_hash = {}
+
+      if should_include_sub_ifds
+        sub_ifds_offsets = raw_exif_data.flat_map(&:sub_ifds).compact
```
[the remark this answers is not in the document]
rcpalacio: good catch! 👍

rcpalacio (commented):
Should be good now! please take a look at my responses :) 
(I'm sorry about the delay... not used to Github's reviews yet, coming from bitbucket )

enricoribelli (approved):

spec/parsers/exif_parser_spec.rb
linkyndy: Alright, got it 👍

linkyndy (approved):
💯

enricoribelli:
🥳

PR #215, opened 2022-09-21
[PREV-144] Loosen/Drop Dependencies, Drop EOL Support and Rename Measurometer Metrics

Kevin-McGonigle:
* Remove the dependencies on Faraday in favour of Net:HTTP (avoids deprecation warnings flooding DD logs).

* Loosen some of the other dependencies.

* Drop support for Ruby <2.7.

* Increment major version.

* Appeasing the updated rubocop styling rules.

* Also including #217 to rename Measurometer metrics to be more DD-friendly.


.github/workflows/main.yml
```diff
@@ -14,9 +14,6 @@ jobs:
       matrix:
         ruby:
```
linkyndy, line 15: I think it would be nice to _support_ the newer versions here as well.

.github/workflows/main.yml
```diff
@@ -14,9 +14,6 @@ jobs:
...
           - 2.7
-          - 2.6
-          - 2.5
-          - jruby
```
linkyndy, line 19: Why drop jruby? 🤔

format_parser.gemspec
```diff
@@ -31,16 +32,16 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
 
   spec.add_dependency 'ks', '~> 0.0'
-  spec.add_dependency 'exifr', '~> 1', '>= 1.3.8'
```
linkyndy, line 34: Do we need to change the version requirements for **all** dependencies?

linkyndy (commented):
Thanks for housekeeping this 🙌

.github/workflows/main.yml
Kevin-McGonigle: From the [faraday_middleware docs](https://github.com/lostisland/faraday_middleware#as-of-v0160-faraday-and-faraday_middleware-no-longer-officially-support-jruby-or-rubinius):

> As of v0.16.0, faraday and faraday_middleware no longer officially support JRuby or Rubinius.

format_parser.gemspec
Kevin-McGonigle: Need? No. But ten birds, one stone - why would we not?

lib/parsers/fdx_parser.rb
```diff
@@ -24,9 +24,9 @@ def xml_check(io)
   def check_for_document_type(file_and_document_type)
     sanitized_data = file_and_document_type.downcase
     if sanitized_data.include?('finaldraft') && sanitized_data.include?('script')
-      return :fdx, :script
```
enricoribelli, line 27: I'd keep the `return`. It might be the "ruby way" to omit it or better, not needing it, but why giving a slight chance to make things unclear (especially to people new to Ruby) ? One more word costs us nothing and gives more clarity imho :)

enricoribelli: Same in another couple of references in this PR

enricoribelli:
Nice big cleanup!! 🤩

lib/parsers/fdx_parser.rb
Kevin-McGonigle: Don't blame me, I agree with you, blame the code style check 🤷🏼‍♂️

enricoribelli:
question just to double confirm: deprecation error is not flooding anymore in DD, right?

lib/parsers/fdx_parser.rb
enricoribelli: really? the linter was complaining about it?

Kevin-McGonigle: As an alternative, I can configure the code style to not check for stuff like this, or any of the other changes that I had to make to appease it. I don't feel strongly either way.

Kevin-McGonigle:
> question just to double confirm: deprecation error is not flooding anymore in DD, right?

I'll deploy to dev and staging today to confirm.

lib/parsers/fdx_parser.rb
enricoribelli: nah it's fine like this.. thanks anyway!

enricoribelli:
> > question just to double confirm: deprecation error is not flooding anymore in DD, right?
> 
> I'll deploy to dev and staging today to confirm.

cool thanks! give me thumbs up after that and I'll ✅  :)

format_parser.gemspec
linkyndy: Because we're narrowing down client support in cases where that might not be necessary.

.github/workflows/main.yml
```diff
@@ -60,15 +57,9 @@ jobs:
...
+          - 3.0
```
linkyndy, line 60: I'd add all the other versions here as well (2.8, 2.9).

.github/workflows/main.yml
linkyndy: I see. It's a pretty big decision to make, only to update a dependency we **chose** to use. This gem is used by many projects out there, some which might be using jruby. I agree dropping support for EOL Ruby versions, but dropping support for something perfectly valid like jruby sounds quite opinionated. Are we sure this is a well-thought decision that we can stand by?

enricoribelli: Well, the thing here is that whether we can assume or not that clients out there use JRuby, our internal used libraries (in this case Faraday) are dropping support for JRuby. Us, per our internal usage, do not need JRuby and to me it makes sense that if our libraries don't support it, we stop supporting it as well.
(Plus I would add a clear message in the changelog about this).

Kevin-McGonigle: +1 to changelog message if we are going ahead with this @enricoribelli, good shout.

As I see it, we have 3 options:
1. Upgrade faraday and drop support for JRuby.
2. Find an alternative to faraday that supports JRuby.
3. Leave everything as-is and just find a way to silence those deprecation warnings (or at least prevent them from reaching DD).

.github/workflows/main.yml
Kevin-McGonigle: There is no 2.8 or 2.9, latest version of Ruby 2 is 2.7.6.

.github/workflows/main.yml
linkyndy: Enrico, an open-sourced project stretches way more than what we currently use internally 🙂 

I agree with everything else, and I'm happy to drop support for jruby if we are all aware of the impact and agree that the best way forward for this gem is without it. It's a conversation that had to happen and I'm happy to see some opinions thrown around.

Since the std HTTP library has likely evolved in all this time, I am wondering whether we couldn't rely on it and thus get rid of an external dependency. Not sure why Faraday was introduced in the first place, but it might be a path worth exploring.

.github/workflows/main.yml
linkyndy: Oh boy, you're right, my mind was somewhere else. Ignore this 😅

.github/workflows/main.yml
Kevin-McGonigle: I'll have a look at the HTTP library and see if it fits our use cases, would definitely be nice if we could drop the dependency altogether 👍🏻

enricoribelli: > Enrico, an open-sourced project stretches way more than what we currently use internally 🙂

For sure! But if Faraday (another open-source library 😛  ) drops JRuby support for its clients (us) we can also look that way with our clients:)

linkyndy: Of course we can, I'm just saying we should be aware that such a decision involves some discussion and alignment.

lib/remote_io.rb
```diff
@@ -73,63 +79,89 @@ def read(n_bytes)
...
-      # we still need the default adapter, more details: https://blog.thecodewhisperer.com/permalink/losing-time-to-faraday
-      faraday.adapter Faraday.default_adapter
-    end
-    response = conn.get(@uri, nil, range: 'bytes=%d-%d' % [range.begin, range.end])
-
-    case response.status
+    response = Net::HTTP.start(@uri.hostname, @uri.port, use_ssl: @uri.scheme == 'https') { |http|
+      http.request_get(@uri, @headers.merge({ 'range' => 'bytes=%d-%d' % [range.begin, range.end] }))
+    }
+    status = Integer(response.code)
```
enricoribelli, line 112: don't need to convert the status code into an integer. You can use directly `Net::HTTPSuccess` built in

spec/remote_fetching_spec.rb
```diff
@@ -124,11 +124,6 @@
   end
 
   it 'sends provided HTTP headers in the request' do
-    # Faraday is required only after calling .parse_http
```
enricoribelli, line 127: happy when we get rid of these magics.......

lib/remote_io.rb
enricoribelli: and other codes too for the code below

lib/remote_io.rb
```diff
@@ -73,63 +79,89 @@ def read(n_bytes)
...
   protected
 
+  REDIRECT_LIMIT = 3
+  UNSAFE_URI_CHARS = %r{[^\-_.!~*'()a-zA-Z\d;/?:@&=+$,\[\]%]}
+
+  # Remove the fragment and escape any unsafe characters in the given URI,
+  # retaining the prior host and scheme if the URI is relative.
+  #
+  # @param uri[String, URI::Generic] The URI to be made safe.
+  def uri=(uri)
```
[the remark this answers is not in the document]
Kevin-McGonigle: a. The intention here is to replicate what Faraday does under the hood when handling redirection. We escape any unsafe characters, and handles any cases where relative paths are returned.

b. We want to do this every time we get redirected - meaning every time we update the `uri` attribute; so it felt appropriate to implement this using a setter so that we can feel confident the URI in use will be valid and pointing to the correct location. It future-proofs it a bit in-case we end up setting the `uri` attribute elsewhere. Feels clean to me, but I'm okay with inlining it or placing it in a simple standalone method if you disagree.

Kevin-McGonigle:
> I have two more remarks, aside from the ones in the review:
> 
> 1. It's great we got rid of Faraday, but it would be good to make sure this doesn't bring any functional/performance penalty. I can imagine there was a good reason Faraday was used, so we should make our due diligence while removing it
> 2. I understand why we've renamed most metrics, however I feel this will break too many things for too many projects for not that much of an advantage. Are we sure we want to go down this path?

(GitHub, why can't I just reply to this comment in a thread? FFS)

1. Sounds like a good idea! I was gonna do this by letting it loose in staging for a while and keeping an eye on DataDog, but I don't love that as an approach... do you know what the best way to go about that?

2. The renamed metrics came in as part of #217, which you approved and [suggested](https://github.com/WeTransfer/format_parser/pull/217#discussion_r977562287) be included as part of this major release. Have you changed your thoughts on it?

linkyndy:
Yes, would be great if they implemented that _feature_.

1. That's a good start. We can draft a 2.0.0.pre release of this gem and test drive that in the wild. We can load test locally, but ultimately it should face the real deal once we are confident enough about the progress.
2. I might have, yes. Thinking more about it, I cannot convince myself there is an actual benefit from causing gem users this amount of pain 😅

lib/remote_io.rb
Kevin-McGonigle: How's that looking to you now?

Kevin-McGonigle:
@linkyndy

1. Cool, I've switched the version to a prerelease.

2. @grdw, could you weigh in on the motivation for renaming the metrics and whether or not it's worth the external headaches it might cause?

grdw:
> could you weigh in on the motivation for renaming the metrics and whether or not it's worth the external headaches it might cause?

It's too not overflow metric systems with a counter per file type, but rather use tags for these kind of situations. Appsignal/Datadog /etc. all support these features. And it's too keep the counters consistent (so, keep it all in snake case).

> [..] and whether or not it's worth the external headaches it might cause?

Do we know if any other companies/people besides WeTransfer use format parser?

lib/remote_io.rb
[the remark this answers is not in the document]
linkyndy: Got it 👌

linkyndy:
As far as I know, there are others using this gem, yes.

linkyndy (approved):
Looks good 👍 (given we test-drive the `pre` version and sort out the metrics concern)

grdw:
> I understand why we've renamed most metrics, however I feel this will break too many things for too many projects for not that much of an advantage. Are we sure we want to go down this path?

Again, if we're not sure how many projects use the actual gem, and moreover use the gem _and_ the metrics, this is not much of a statement 💭. As long as we described in the CHANGELOG that we changed it, then we should be good as far as I'm concerned. 

We can also keep the metrics _as is_; with the funny snake-camel-case, fine with me as well 🤷. But then what? How will we ever change it in the future?

enricoribelli (approved):

linkyndy:
Fair point. Feels like we've reached a consensus on this one. Let's 🚀  then!

PR #217, opened 2022-09-22
Rename metrics

grdw:
This PR renames the metrics to make sure they're all:

- Written in underscore / snake case (for consistency)
- Uses tags where appropriate (this would prevent metric systems from being flooded with similar looking metric names)

### DIFF
--- a/lib/care.rb
+++ b/lib/care.rb
@@ -170,10 +170,10 @@ class Care
     # @param io[IO] the IO to read from
     # @param page_i[Integer] which page (zero-based) to read
     def read_page(io, page_i)
-      Measurometer.increment_counter('format_parser.parser.Care.page_reads_from_upsteam', 1)
+      Measurometer.increment_counter('format_parser.parser.care.page_reads_from_upsteam', 1)
 
       io.seek(page_i * @page_size)
-      read_result = Measurometer.instrument('format_parser.Care.read_page') { io.read(@page_size) }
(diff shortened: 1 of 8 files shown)


lib/format_parser/version.rb
```diff
@@ -1,3 +1,3 @@
 module FormatParser
-  VERSION = '1.6.0'
+  VERSION = '1.6.1'
```
grdw, line 2: ☝️ I'm kind of unclear if this is the right version bump. In a way, changing metric names isn't a breaking change? But it can be considered as such _if_ people make use of the metrics from this gem in their own separate systems, and this causes missing data. I'm not sure if this has the right weight, but please correct me if this needs a more serious bump.

pcriv (approved):

lib/format_parser/version.rb
grdw: Ah fair enough 👍 . Shall I change my PR base-branch to #[kev/prev-144/upgrade-dependencies-drop-eol](https://github.com/WeTransfer/format_parser/tree/kev/prev-144/upgrade-dependencies-drop-eol) or? 

Incl. @Kevin-McGonigle wdyt?

Kevin-McGonigle (approved):
Feel free to merge this to [my branch](https://github.com/WeTransfer/format_parser/tree/kev/prev-144/upgrade-dependencies-drop-eol) when you're happy it's gotten enough approval 👍🏻

pcriv (approved):

linkyndy (approved):

PR #218, opened 2022-10-03
Add support for .ARW files

RubenWolff-wt:
- split .arw files out of the existing tiff_parser into their own arw_parser
- fix some bugs around some .ARW files having a padded make tag (which was not being read correctly previously)
- added a lot more different test files

### DIFF
--- a/lib/parsers/tiff_parser.rb
+++ b/lib/parsers/tiff_parser.rb
@@ -6,7 +6,6 @@ class FormatParser::TIFFParser
   MAGIC_BE = [0x4D, 0x4D, 0x0, 0x2A].pack('C4')
   HEADER_BYTES = [MAGIC_LE, MAGIC_BE]
   TIFF_MIME_TYPE = 'image/tiff'
-  ARW_MIME_TYPE = 'image/x-sony-arw'
 
   def likely_match?(filename)
     filename =~ /\.tiff?$/i
@@ -28,20 +27,21 @@ class FormatParser::TIFFParser
     exif_data = exif_from_tiff_io(io)
     return unless exif_data
 
+    # arw files are now handled in separate arw_parser 
+    return if arw?(exif_data)
+
     w = exif_data.width || exif_data.pixel_x_dimension
     h = exif_data.height || exif_data.pixel_y_dimension
 
-    format = arw?(exif_data) ? :arw : :tif
-    mime_type = arw?(exif_data) ? ARW_MIME_TYPE : TIFF_MIME_TYPE
     FormatParser::Image.new(
-      format: format,
+      format: :tif,
       width_px: w,
       height_px: h,
       display_width_px: exif_data.rotated? ? h : w,
       display_height_px: exif_data.rotated? ? w : h,
       orientation: exif_data.orientation_sym,
       intrinsics: {exif: exif_data},
-      content_type: mime_type,
+      content_type: TIFF_MIME_TYPE,
     )
   rescue EXIFR::MalformedTIFF
     nil
@@ -54,8 +54,9 @@ class FormatParser::TIFFParser
 
   # Similar to how exiftool determines the image type as ARW, we are implementing a check here
   # https://github.com/exiftool/exiftool/blob/e969456372fbaf4b980fea8bb094d71033ac8bf7/lib/Image/ExifTool/Exif.pm#L929
+  # Change in version 1.7.0 to use start_with for make matching due to some sample images having a padded make
   def arw?(exif_data)
-    exif_data.compression == 6 && exif_data.new_subfile_type == 1 && exif_data.make == 'SONY'
+    exif_data.compression == 6 && exif_data.new_subfile_type == 1 && exif_data.make&.start_with?('SONY')
   end
 
   FormatParser.register_parser new, natures: :image, formats: :tif
--- a/lib/parsers/tiff_parser.rb
+++ b/lib/parsers/tiff_parser.rb
@@ -27,7 +27,6 @@ class FormatParser::TIFFParser
     exif_data = exif_from_tiff_io(io)
     return unless exif_data
 
-    # arw files are now handled in separate arw_parser
     return if arw?(exif_data)
 
     w = exif_data.width || exif_data.pixel_x_dimension
@@ -54,7 +53,6 @@ class FormatParser::TIFFParser
 
   # Similar to how exiftool determines the image type as ARW, we are implementing a check here
   # https://github.com/exiftool/exiftool/blob/e969456372fbaf4b980fea8bb094d71033ac8bf7/lib/Image/ExifTool/Exif.pm#L929
-  # Change in version 1.7.0 to use start_with for make matching due to some sample images having a padded make
   def arw?(exif_data)
     exif_data.compression == 6 && exif_data.new_subfile_type == 1 && exif_data.make&.start_with?('SONY')
   end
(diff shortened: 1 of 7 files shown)


spec/parsers/arw_parser_spec.rb
```diff
@@ -0,0 +1,103 @@
+require 'spec_helper'
+
+describe FormatParser::ARWParser do
+  describe 'matches filenames with valid extensions' do
+    filenames = ['raw_file', 'another raw file', 'and.another', 'one-more']
+    extensions = ['.arw', '.Arw', '.aRw', '.arW', '.ARw', '.ArW', '.aRW', '.ARW']
+    filenames.each do |filename|
```
rcpalacio, line 7: Maybe using rspec's `include_examples` and `shared_examples` could make this a bit cleaner :)

RubenWolff-wt: I'm not familiar with shared_examples (I had a quick look and it looks like it can factor out repeated test code especially for a type or module https://relishapp.com/rspec/rspec-core/v/2-14/docs/example-groups/shared-examples) but what would you like me to clean up here?

The cross join between the 2 collections to get the test file names? As the assertion is different between the 2 tests factoring out the entire test code seems more work than it would save but maybe I'm miss-understanding what you are suggesting.

rcpalacio: We don't use it a lot throughout the code base, but `aac_parser_spec.rb` and a few other places benefit from it.

Even if you don't use these examples it across multiple describe blocks, it still gives you a descriptive way to setup examples for your multiple `it` blocks. Using it, you don't need to explicitly call `.each` and it makes it easier to come up with a `describe` block that won't look like an `it`.

The way you have it is definitely fine. I bring it up because I found it neat when I saw the guys using `shared_examples`.

lib/parsers/tiff_parser.rb
```diff
@@ -54,8 +54,9 @@ def cr2?(io)
 
   # Similar to how exiftool determines the image type as ARW, we are implementing a check here
   # https://github.com/exiftool/exiftool/blob/e969456372fbaf4b980fea8bb094d71033ac8bf7/lib/Image/ExifTool/Exif.pm#L929
+  # Change in version 1.7.0 to use start_with for make matching due to some sample images having a padded make
```
linkyndy, line 57: Not sure whether we need this comment 🤔

lib/parsers/tiff_parser.rb
```diff
@@ -28,20 +27,21 @@ def call(io)
     exif_data = exif_from_tiff_io(io)
     return unless exif_data
 
+    # arw files are now handled in separate arw_parser
```
linkyndy, line 30: Also this comment seems a bit redundant, in my opinion.

RubenWolff-wt: 👍 I'll remove it

lib/parsers/tiff_parser.rb
RubenWolff-wt: 👍 I'll remove it. Wasn't sure if we wanted to indicate existing functionality changes back to versions but I doubt it's very useful.

spec/parsers/arw_parser_spec.rb
RubenWolff-wt: Thanks for explaining it and pointing me at some samples.

I've added some extra tests and cleaned up some of the existing likely_match tests to use shared_examples.

I also added some error cases as shared examples to confirm that other file formats do not get parsed.

CostinTanasoiu (approved):
It looks good! Congrats!

spec/parsers/arw_parser_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+  end
+
+  shared_examples 'no likely_match for file' do |filename_with_extension|
+    it "does not match '#{filename_with_extension}'" do
+      expect(subject.likely_match?(filename_with_extension)).to be_falsey
+    end
+  end
+
+  describe 'likely_match' do
```
rcpalacio, line 16: I get the idea behind building all these examples programatically to be more extensive. It feels a bit odd to have a "describe" section holding this though.
Maybe this would fit better on a `before` section? or maybe we can cover this with a smaller set of examples but list them one-by-one? What do you think?

RubenWolff-wt: I personally prefer programatically creating test cases over a long list of manually created test cases especially when I want to quickly create combinations for a simple method like this.

There are fewer strings so there is less chance of a mistake or repeated tests and the loops are fairly self explanatory in my opinion. The outcome of this is human readable so understanding the test outcome is not affected by the loop.

`
bundle exec rspec spec/parsers/arw_parser_spec.rb --format documentation

Randomized with seed 41432

FormatParser::ARWParser
  parses Sony ARW fixtures as arw format file
    is able to parse RAW_SONY_SLTA55V.ARW
    is able to parse RAW_SONY_A900.ARW
    should fail to parse JPG
    should fail to parse CR2
    should fail to parse PNG
    is able to parse RAW_SONY_DSC-RX100M2.ARW
    is able to parse RAW_SONY_A700.ARW
    is able to parse RAW_SONY_NEX7.ARW
    should fail to parse NEF
    is able to parse RAW_SONY_ILCE-7RM2.ARW
    should fail to parse TIFF
    is able to parse RAW_SONY_A100.ARW
  likely_match
    matches 'and.another.Arw'
    matches 'another raw file.ARW'
    matches 'one-more.aRW'
    does not match 'another raw file.tiff'
    matches 'raw_file.ARW'
    matches 'raw_file.arW'
    does not match 'raw_file.tiff'
    matches 'one-more.arw'
    does not match 'raw_file.new'
    matches 'raw_file.aRw'
    matches 'another raw file.arW'
    matches 'one-more.aRw'
    matches 'another raw file.ArW'
    matches 'and.another.aRW'
    matches 'and.another.ArW'
    does not match 'raw_file.cr2'
    does not match 'one-more.jpeg'
    matches 'raw_file.ArW'
    does not match 'and.another.jpeg'
    matches 'one-more.Arw'
    matches 'one-more.ArW'
    matches 'one-more.arW'
    matches 'and.another.arw'
    matches 'one-more.ARw'
    does not match 'raw_file.jpeg'
    does not match 'one-more.new'
    does not match 'and.another.tiff'
    does not match 'another raw file.new'
    matches 'raw_file.Arw'
    matches 'raw_file.ARw'
    matches 'another raw file.ARw'
    does not match 'another raw file.jpeg'
    matches 'and.another.ARw'
    matches 'raw_file.arw'
    matches 'another raw file.Arw'
    matches 'another raw file.arw'
    matches 'one-more.ARW'
    does not match 'and.another.new'
    matches 'and.another.aRw'
    matches 'and.another.ARW'
    matches 'another raw file.aRw'
    does not match 'one-more.cr2'
    does not match 'one-more.tiff'
    does not match 'another raw file.cr2'
    matches 'and.another.arW'
    does not match 'and.another.cr2'
    matches 'raw_file.aRW'
    matches 'another raw file.aRW'
`

linkyndy: I think it makes sense, for **this** use-case.

I would only argue whether we need to test all those permutations of valid extensions and filenames, as it feels a bit aside from the actual logic that is tested. If it was me, I would test `anything.` + `(arw|tif|cr2|anything-else)`.

RubenWolff-wt: Yeah the number of test cases is a be a bit overkill (I copied the initial set of all case permutations of extensions from the aac tests).

As we are using regex I do think there is some worth in at least including some different casing in the extension as well as adding some odd file names (eg including a period) to ensure there are no unexpected outcomes.

eg someone accidentally removing the i from the regex would be caught by including .ARW.

rcpalacio (approved):
Looks good!

PR #219, opened 2022-10-18
[NO JIRA] Clear Authorization Header on Redirect

Kevin-McGonigle:
When Faraday would follow a redirect, by default, it would clear the `Authorization` header if the redirect location had a different host than the original request. This replicates that behaviour, the absence of which was causing S3 to respond with `InvalidArgument - Only one auth mechanism allowed` on Storm redirects.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
         run: bundle exec rubocop
   test:
     name: Specs
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ env:
 jobs:
   lint:
     name: Code Style
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
--- a/spec/remote_io_spec.rb
+++ b/spec/remote_io_spec.rb
@@ -251,6 +251,7 @@ describe FormatParser::RemoteIO do
   end
 
   # 5XX
+
   (500..599).each do |code|
     context "when the response status code is #{code}" do
       it 'raises an error' do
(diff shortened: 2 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ env:
 jobs:
   lint:
     name: Code Style
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
Kevin-McGonigle, line 11: Was getting blocked by a brownout since [18.04 is being deprecated](https://github.com/actions/runner-images/issues/6002), this works though 😄

Kevin-McGonigle:
@linkyndy took it one step further and tidied up the test cases as a whole, and added some extra tests for previously missed cases.

linkyndy (commented):
I would highly recommend using [webmock](https://thoughtbot.com/blog/stubbing-and-setting-expectations-on-http-requests-is-now-easy-with-webmock) instead of stubbing `Net::HTTP` heavily.

spec/remote_io_spec.rb
```diff
@@ -3,187 +3,343 @@
...
-      a_hash_including('range' => 'bytes=100-199')
-    )
-
-    rio = described_class.new(url)
-    rio.seek(100)
-
-    expect(rio.read(100)).to be_nil
+  # 3XX
+
+  context 'when the response code is 301 (Moved Permanently), 302 (Found), 303 (See Other), 307 (Temporary Redirect) or  308 (Permanent Redirect)' do
```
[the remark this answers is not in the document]
Kevin-McGonigle: By that, do you mean you want a separate context and test for _every single response code_ where we check for the same behaviour each time?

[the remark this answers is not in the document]
linkyndy: Just moving the loop from the tests to above the examples/contexts, which is what you did 👍

spec/spec_helper.rb
```diff
@@ -30,3 +34,5 @@ def fixtures_dir
     end
   end
 end
+
+RSpec::Matchers.define_negated_matcher :excluding, :include
```
linkyndy, line 38: I think we don't need this one anymore, right?

linkyndy (approved):

spec/spec_helper.rb
Kevin-McGonigle: Good spot!

PR #222, opened 2022-11-03
[PREV-165] CR3 Support

Kevin-McGonigle:


lib/parsers/moov_parser.rb
```diff
@@ -6,19 +6,21 @@ class FormatParser::MOOVParser
...
   # https://tools.ietf.org/html/rfc4337#section-2
   # There is also video/quicktime which we should be able to capture
   # here, but there is currently no detection for MOVs versus MP4s
+  CR3_MIME_TYPE = 'image/x-canon-cr3'
   MP4_AU_MIME_TYPE = 'audio/mp4'
   MP4_MIXED_MIME_TYPE = 'video/mp4'
 
   def likely_match?(filename)
-    filename =~ /\.(mov|m4a|ma4|mp4|aac|m4v)$/i
+    filename =~ /\.(aac|cr3|m4a|m4v|ma4|mov|mp4)$/i
```
RubenWolff-wt, line 23: I know this is not a change you made but I am somewhat surprised to see aac here. It's not in FTYP_MAP above and there is an entirely separate aac parser IIRC.

lib/parsers/moov_parser.rb
```diff
@@ -161,5 +165,36 @@ def parse_sample_description_atom(decoder, atom_tree)
     end
   end
 
-  FormatParser.register_parser new, natures: :video, formats: FTYP_MAP.values, priority: 3
+  def cr3(decoder, atom_tree, fallback_width, fallback_height)
```
RubenWolff-wt, line 168: I'm not sure if it's correct to have this in the moov parser as opposed to using a separate parser. To me it seems like the logic is quite separate and it's just reusing the atom decoder (not that I know much aout the file format so maybe cr3 files are truly moov files 🤷).

For example the fallback width and height here seem to be very video specific using decoder.find_video_trak_atom which does not seem like it should be used for an image.

lib/parsers/moov_parser.rb
```diff
@@ -55,17 +58,18 @@ def call(io)
...
       FormatParser::Audio.new(
-        format: format_from_moov_type(file_type),
+        format: format,
         media_duration_seconds: media_duration_s,
         content_type: MP4_AU_MIME_TYPE,
         intrinsics: atom_tree,
       )
+    elsif format == :cr3
+      cr3(decoder, atom_tree, width, height)
     else
```
RubenWolff-wt, line 70: Again not an issue with your code at all but would this not also run if no format was found? Say if the file_type value is something unknown would we just always default it to a video with format nil as oposed to making it clear we do not handle the file format by returning nil?

lib/parsers/moov_parser/decoder.rb
```diff
@@ -4,6 +4,7 @@
 # To know more about Atoms: https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html
 class FormatParser::MOOVParser::Decoder
   include FormatParser::IOUtils
+  include FormatParser::EXIFParser
```
RubenWolff-wt, line 7: Feels odd to add an image specific dependency (exif parser) to a moov specific decoder. Similar to some other comments I added it feel a bit like there issome core common functionality regarding atoms and then 2 quite specific uses of this (one the existing moov logic and the other the new image related logic).

Would it be possible to have a common AtomDecoder (usable anywhere) then this moov decoder that wraps the atomdecoder with moov specific logic and a separate CR3 decoder which wraps the AtomDecoder with cr3 specific logic?

lib/parsers/moov_parser.rb
Kevin-McGonigle: Agreed, I'll take it out.

lib/parsers/moov_parser.rb
Kevin-McGonigle: It looks like we default to `:mov` if the `file_type` value isn't something we recognise, so having it default to a video is appropriate based on that. Whether or not that's a good approach is another thing, and I would probably lean "no" 😅 But we have operated with it that way since 2018, apparently, and probably warrants its own story to sort it out.

lib/parsers/moov_parser/decoder.rb
Kevin-McGonigle: Yeah I like that as an idea, I'll give it a go!

lib/parsers/moov_parser.rb
Kevin-McGonigle: CR3 files do adapt the [ISO base media file format](https://en.wikipedia.org/wiki/ISO_base_media_file_format), which is where the `moov` container comes from, and only includes a minor extension on the traditional format to include some Canon-specific metadata - so I thought it was appropriate, but I see the argument for the contrary (especially with your `AtomDecoder` comment in mind).

To your second point, CR3 files _do_ (oddly) have video `trak` atoms, despite them describing single frames of the same image with different properties (dimensions, dual-pixels, compression, etc.). Agree that it's weird, but it does apply and would serve as an appropriate fallback if we can't find the CMT1 atom for some reason to parse the EXIF there.

spec/parsers/cr3_parser_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'spec_helper'
+
+describe FormatParser::CR3Parser do
```
CostinTanasoiu, line 3: Could you add some negative tests as well? I.e. test that some files can't be parsed. 

Also, you can add some tests for the `likely_match?` method as well, that check for case insensitivity, as well as some negative scenarios. You can check the `aac_parser_spec.rb` file, and get inspiration from the following tests:
- "should match filenames with valid AAC extensions"
- "should not match filenames with invalid AAC extensions"

lib/parsers/cr3_parser/decoder.rb
```diff
@@ -0,0 +1,34 @@
+require 'parsers/iso_base_file_format/decoder'
+
+class FormatParser::CR3Parser::Decoder < FormatParser::Decoder
```
CostinTanasoiu, line 3: Tests?

CostinTanasoiu (changes requested):
A few more things:
- Update the gem version
- Update the `readme.md` file and add your CR3 fixture sources.

lib/parsers/cr3_parser/decoder.rb
```diff
@@ -0,0 +1,34 @@
...
+      [fields, nil]
+    else
+      [nil, nil]
+    end
+  end
+
+  def uuid(size)
+    usertype = read_bytes(16).unpack('H*').first
+    fields = { usertype: usertype }
+    children = if usertype == '85c0b687820f11e08111f4ce462b6a48'
```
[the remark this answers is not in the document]
Kevin-McGonigle: It's the Canon-specific metadata container, see [here](https://github.com/lclevy/canon_cr3)

lib/parsers/cr3_parser/decoder.rb
Kevin-McGonigle: The tests in the cr3_parser_spec cover the functionality here.

spec/parsers/cr3_parser_spec.rb
Kevin-McGonigle: Got some examples of negative tests? Do you mean corrupted files, for example? That's not something I've seen tested for the majority of other formats.

Kevin-McGonigle: Added in tests for the escape cases (when the file doesn't match the CR3 definition and when the CMT1 atom can't be found) and for the extension matching.

RubenWolff-wt (commented):
LGTM overall but one of the new tests is still failing.

Just out of interest did you write the entire base encoder from scratch or did you use something for this?

Kevin-McGonigle:
> LGTM overall but one of the new tests is still failing.
> 
> Just out of interest did you write the entire base encoder from scratch or did you use something for this?

Yeah, it's actually an existing test and an annoying one since it's only JRuby failing and only on the *purposely invalid* file that's being picked up in the glob.

Kind of a halfway-house: I wrote it from scratch based on the pseudocode in the standard [here](https://b.goeswhere.com/ISO_IEC_14496-12_2015.pdf). Almost none of the atoms are of interest to us, but with it being a public repo I figured it'd be best to implement in full.

Re: @RubenWolff-wt

RubenWolff-wt:
> > LGTM overall but one of the new tests is still failing.
> > Just out of interest did you write the entire base encoder from scratch or did you use something for this?
> 
> Yeah, it's actually an existing test and an annoying one since it's only JRuby failing and only on the _purposely invalid_ file that's being picked up in the glob.
> 
> Kind of a halfway-house: I wrote it from scratch based on the pseudocode in the standard [here](https://b.goeswhere.com/ISO_IEC_14496-12_2015.pdf). Almost none of the atoms are of interest to us, but with it being a public repo I figured it'd be best to implement in full.
> 
> Re: @RubenWolff-wt

Noice. How did you find that doc? If it's not easily accessible from the link you added to the common decoder already I think it's worth adding it there as well to help any future devs (not that I expect many people to need to look into this in detail)

lib/parsers/iso_base_file_format/decoder.rb
```diff
@@ -0,0 +1,1044 @@
+# This class provides generic methods for parsing file formats based on QuickTime-style "atoms", such as those seen in
+# the ISO base media file format (ISO/IEC 14496-12), a.k.a MPEG-4, and those that extend it (MP4, CR3, HEIF, etc.).
+#
+# For more information on atoms, see https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap1/qtff1.html
+
+class FormatParser::Decoder
```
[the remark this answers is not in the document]
CostinTanasoiu: As you mentioned in another comment, you implemented a lot of those other atoms because it's a public repo so it would be of interest to others. I think the fact that we added support for those atoms in a public repo warrants having tests for it. 

It's a massive undertaking though, so maybe you could remove what we don't use for now, and only add them when we can actually provide tests for them... I'm just uncomfortable with publishing untested code to a public open-source project...

spec/parsers/cr3_parser_spec.rb
CostinTanasoiu: Nice work! I'm happy with those tests.

lib/parsers/iso_base_file_format/decoder.rb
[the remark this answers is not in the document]
Kevin-McGonigle: Aight, I'll comment out what's not necessary for CR3 and we can reintroduce them down the line.

CostinTanasoiu (approved):
Good work!

README.md
```diff
@@ -12,29 +12,37 @@ and [dimensions,](https://github.com/sstephenson/dimensions) borrowing from them
 
 ## Currently supported filetypes:
 
-* TIFF
```
linkyndy, line 15: Thanks for ordering these so neatly 🙏

lib/io_utils.rb
```diff
@@ -27,29 +27,59 @@ def safe_skip(io, n)
...
+    [n >> 30, n & 0x3fffffff]
   end
 
-  # 'n' is the number of bytes to read
   def read_string(n)
     safe_read(@buf, n)
   end
 
+  # 'n' is the number of bytes to read
+  def read_bytes(n)
```
linkyndy, line 71: Do we need two methods for this? If it's just for added sugar, you could `alias_method` one of them.

lib/io_utils.rb
```diff
@@ -27,29 +27,59 @@ def safe_skip(io, n)
...
+    safe_read(@buf, n)
+  end
+
+  def skip_bytes(n)
+    safe_skip(@buf, n)
+  end
+
+  def skip_bytes_then(n)
```
linkyndy, line 79: You could also have just one method for this and `yield if block_given?` inside it.

lib/parsers/cr3_parser.rb
```diff
@@ -0,0 +1,49 @@
...
+        atom_tree: atom_tree,
+        exif: cmt1_atom.fields,
+      },
+    )
+  end
+
+  private
+
+  def matches_cr3_definition?
+    matches = skip_bytes_then(4) { read_string(8) } == 'ftypcrx '
```
linkyndy, line 43: I would extract the magic string to a constant.

linkyndy, line 43: Maybe I'm missing something, but how can skipping 4 bytes be equal to a string? If I'm not, maybe we could improve the semantics of this a bit, so it's less confusing.

lib/parsers/cr3_parser.rb
```diff
@@ -0,0 +1,49 @@
+class FormatParser::CR3Parser
+  include FormatParser::IOUtils
+  require_relative 'cr3_parser/decoder'
```
linkyndy, line 3: Do we _need_ to require inside the class definition?

lib/parsers/iso_base_media_file_format/decoder.rb
```diff
@@ -0,0 +1,1039 @@
...
+      def parse_atom
+        position = @buf.pos
+
+        size = read_int_32
+        type = read_string(4)
+        size = read_int_64 if size == 1
+        body_size = size - (@buf.pos - position)
+        next_atom_position = position + size
+
+        if self.class::ATOM_PARSERS.include?(type)
```
linkyndy, line 165: Why not just:

```suggestion
        if ATOM_PARSERS.include?(type)
```

lib/parsers/iso_base_media_file_format/decoder.rb
```diff
@@ -0,0 +1,1039 @@
+# This class provides generic methods for parsing file formats based on QuickTime-style "atoms", such as those seen in
+# the ISO base media file format (ISO/IEC 14496-12), a.k.a MPEG-4, and those that extend it (MP4, CR3, HEIF, etc.).
+#
+# For more information on atoms, see https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap1/qtff1.html
+# or https://b.goeswhere.com/ISO_IEC_14496-12_2015.pdf.
+#
+# The vast majority of the methods have been commented out here. This decision was taken to expedite the release of
```
linkyndy, line 7: I would mark this with a `# TODO:`, so it's more visible that we (or anyone else) expect to get back to this, sooner or later.

lib/parsers/moov_parser.rb
```diff
@@ -18,7 +18,7 @@ class FormatParser::MOOVParser
   MP4_MIXED_MIME_TYPE = 'video/mp4'
 
   def likely_match?(filename)
-    filename =~ /\.(mov|m4a|ma4|mp4|aac|m4v)$/i
```
linkyndy, line 21: What happened with `aac`? 🤔

lib/parsers/moov_parser/decoder.rb
```diff
@@ -77,16 +77,63 @@ def find_video_trak_atom(atoms)
...
+  # we know are permitted to have leaf/branch atoms within itself,
+  # and will attempt to recover the data fields for leaf atoms
+  #
+  # @param [IO] io
+  # @param [Integer] max_read
+  # @param [Array<String>] current_branch
+  # @return [Array<Atom>]
+  def extract_atom_stream(io, max_read, current_branch = [])
+    initial_pos = io.pos
+    atoms = []
```
linkyndy, line 90: You could also use `each_with_object` to shorten declaration+block+return to just block, if it makes more sense to you 😊

spec/format_parser_spec.rb
```diff
@@ -177,14 +177,16 @@
...
         [:audio, :image],
-        [:cr2, :dpx, :fdx, :flac, :gif, :jpg, :mov, :mp4, :m4a, :mp3, :mpg, :mpeg, :ogg, :png, :tif, :wav]
+        [:cr2, :cr3, :dpx, :fdx, :flac, :gif, :jpg, :mov, :mp4, :m4a, :mp3, :mpg, :mpeg, :ogg, :png, :tif, :wav]
       )
 
       expect(parsers.map { |parser| parser.class.name }).to eq([
         'FormatParser::GIFParser',
         'Class',
         'FormatParser::PNGParser',
+        'FormatParser::MOOVParser',
```
linkyndy, line 187: Curious, why wasn't this parser here before? 🤔

spec/parsers/iso_base_media_file_format/decoder_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+        expect(nmhd_atom.children).to eq([])
+      end
+    end
+  end
+end
+
+describe FormatParser::ISOBaseMediaFileFormat::Decoder::Atom do
+  context 'when initialized' do
+    context 'without fields and/or children' do
+      subject { FormatParser::ISOBaseMediaFileFormat::Decoder::Atom.new('foo', 0, 0) }
```
linkyndy, line 132: You can make things a bit more bearable with:

```suggestion
      subject { described_class.new('foo', 0, 0) }
```

lib/io_utils.rb
Kevin-McGonigle: Yeah I wanted it just to communicate intention a bit more clearly. Didn't know `alias_method` was a thing, awesome!

lib/io_utils.rb
Kevin-McGonigle: Ooh that's cool, thanks fot the tip!

lib/parsers/iso_base_media_file_format/decoder.rb
Kevin-McGonigle: Doing it this way means that it continues to function with extending class (such as the CR3 decoder). This way, all that extending classes have to do is add to its own ATOM_PARSERS, like [so](https://github.com/WeTransfer/format_parser/pull/222/files#diff-8e8d366860cab7abfad65caebdb7891360815ec2013a1cad34d4efdb7ee81c06R8-R10).

lib/parsers/moov_parser.rb
Kevin-McGonigle: Ruben raised [here](https://github.com/WeTransfer/format_parser/pull/222#discussion_r1013725910) that it doesn't really make any sense for it to be here. It's not in the FTYP map and Costin wrote a [dedicated parser](https://github.com/WeTransfer/format_parser/blob/master/lib/parsers/aac_parser.rb) for it a short while ago.

linkyndy: Alright, seems reasonable. Let's just make sure we're not altering any of the unseen with this.

lib/parsers/moov_parser/decoder.rb
Kevin-McGonigle: This is untouched from its previous implementation, I think it's just been relocated within the file as a side effect of old changes that I replaced with this dedicated ISO decoder approach. Idea is that this will be replaced down the line anyway, so I don't see a need to mess with it - good tip though!

spec/format_parser_spec.rb
Kevin-McGonigle: It [was](https://github.com/WeTransfer/format_parser/pull/222/files/f08a399d1d97d5674e146232c542608b8b7130aa#diff-9a6e94baf6d000b5721f4990cf2d55457d60c94614851c6fdcc83dc455da772cL164) only being registered with `:video` as a nature, but that's not representative since it returns `Audio` for M4A files, so I a fixed that up.

spec/parsers/iso_base_media_file_format/decoder_spec.rb
Kevin-McGonigle: Love it! Thanks!

lib/parsers/moov_parser/decoder.rb
Kevin-McGonigle: In fact, I'll revert all changes to the MOOV decoder, to be extra-sure.

lib/parsers/iso_base_media_file_format/decoder.rb
linkyndy: Ah, got it. It is a bit subtle, seems more suitable to leverage inheritance here. But I'll leave it up to you.

Kevin-McGonigle: Tried doing it with standard inheritance, but I think I'm too much of a ruby n00b and this was what I found online as the suggested way to modify an inherited constant and have it be picked up by an inherited method without unintended changes to the base class or other inheriting classes. Played around with it a good bit and this was the way that worked. If you've got a suggestion for how to make this more intuitive then I'm definitely all ears!

lib/parsers/cr3_parser.rb
Kevin-McGonigle: Yep, otherwise we get the following in FormatParser::CR3Parser::Decoder
'''
Failure/Error:
  class FormatParser::CR3Parser::Decoder < FormatParser::ISOBaseMediaFileFormat::Decoder
    include FormatParser::EXIFParser

    protected

    ATOM_PARSERS = ATOM_PARSERS.merge({
      'CMT1' => :cmt1
    })
    CANON_METADATA_CONTAINER_UUID = '85c0b687820f11e08111f4ce462b6a48'


NameError:
  uninitialized constant FormatParser::CR3Parser

  class FormatParser::CR3Parser::Decoder < FormatParser::ISOBaseMediaFileFormat::Decoder
                    ^^^^^^^^^^^
  Did you mean?  FormatParser::CR2Parser
                 FormatParser::ARWParser
'''

lib/parsers/iso_base_media_file_format/decoder.rb
linkyndy: We can have a look together next week, but this works as well, so don't let it block you 😊

lib/parsers/cr3_parser.rb
linkyndy: I meant, wouldn't it make more sense to put it at the top of the file?

Kevin-McGonigle: Yes, that's what we get when it's at the top of the file. I think it's analagous to a circular import in Python if it's at the top of the file - if it's at the top, it tries to require the decoder, but the decoder then tries to declare itself under `FormatParser::CR3Parser` which hasn't yet been declared because it's waiting to require the decoder which hasn't yet been declared because it's waiting for the parser etc.

linkyndy: Got it. Normally, you'd require things from a central place, so you wouldn't run into too much trouble doing it all over the place. But that's a topic for another discussion ✌️

PR #223, opened 2022-11-03
2.0.0 Release

Kevin-McGonigle:
The pre-release has been deployed in Peek prod all day, all looks good! A very small job latency improvement (~10% quicker on average) and no discernible impact on parsing error rates. The big win is the reduction in log volume - without the deprecation warnings, we get ~half the number of error log messages that we otherwise would.

Screenshot 2022-11-03 at 16 06 39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-## 2.0.0.pre.2 (Prerelease)
-* Drop support for Ruby `<2.7`.
+## 2.0.0
+* Drop explicit support for Ruby `<2.7`.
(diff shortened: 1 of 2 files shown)


RubenWolff-wt (approved):
LGTM - Do we need some kind of documentation outlining what the breaking changes from v 1.x is

Kevin-McGonigle:
> LGTM - Do we need some kind of documentation outlining what the breaking changes from v 1.x is

@RubenWolff-wt Good catch! Was already there but forgot to update the heading.

RubenWolff-wt (approved):

PR #224, opened 2022-11-07
chore: drop ks dependency

idanci:
Ruby 2.6 (our minimum required version) supports keyword_init for structs out of the box, we can drop the `ks` dependency now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,6 +60,7 @@ jobs:
     strategy:
       matrix:
         ruby:
+          - 2.6
           - 2.7
           - 3.0
           - 3.1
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -13,6 +13,7 @@ jobs:
     strategy:
       matrix:
         ruby:
+          - 2.6
```
idanci, line 16: If we commit to support jruby, then we have to ensure compatibility with 2.6

Kevin-McGonigle (changes requested):
Looks good, but can you add 2.6 to the `tests` job too? Right now it's just on code style.

Kevin-McGonigle (approved):
Thanks for the housekeeping! 🚀

PR #225, opened 2022-11-28
[CORE-199] Exclude CR3 from MOOV definition matching.

Kevin-McGonigle:
CR3 files are being mididentified as MOOV files because they fit the defintion matching function and the CR3 parser has lower priority.


linkyndy (approved):

RubenWolff-wt (approved):
LGTM

PR #226, opened 2022-12-05
[CORE-88] RW2 Support

Kevin-McGonigle:
Add support for Panasonic RAW formats (primarily RW2, but also get RAW and RWL for free - all EXIF-based with the same mimetype).

### DIFF
--- a/lib/parsers/rw2_parser.rb
+++ b/lib/parsers/rw2_parser.rb
@@ -1,0 +1,61 @@
+require_relative 'exif_parser'
+
+class FormatParser::RW2Parser
+  include FormatParser::IOUtils
+  include FormatParser::EXIFParser
+
+  def likely_match?(filename)
+    filename =~ /\.(rw2|raw|rwl)$/i
+  end
+
+  def call(io)
+    @buf = FormatParser::IOConstraint.new(io)
+
+    return unless matches_rw2_definition?
+
+    @buf.seek(0)
+    exif = exif_from_tiff_io(@buf)
+    return unless exif
+
+    # RW2 doesn't use the standard EXIF width and height tags (🤷🏻). We can compute them from the sensor
+    # top/bottom/left/right border tags. See https://exiftool.org/TagNames/PanasonicRaw.html for more.
+    w = sensor_border(exif, :right) - sensor_border(exif, :left)
+    h = sensor_border(exif, :bottom) - sensor_border(exif, :top)
+
+    FormatParser::Image.new(
+      format: :rw2,
+      width_px: w,
+      height_px: h,
+      display_width_px: exif.rotated? ? h : w,
+      display_height_px: exif.rotated? ? w : h,
+      orientation: exif.orientation_sym,
+      intrinsics: { exif: exif },
+      content_type: PANASONIC_RAW_MIMETYPE,
+    )
+  rescue EXIFR::MalformedTIFF
+    nil
+  end
+
+  private
+
+  PANASONIC_RAW_MIMETYPE = 'image/x-panasonic-raw'
+  RW2_MAGIC_BYTES = [0x49, 0x49, 0x55, 0x0, 0x18, 0x0, 0x0, 0x0].pack('C8')
+  RAW_RWL_MAGIC_BYTES = [0x49, 0x49, 0x55, 0x0, 0x08, 0x0, 0x0, 0x0].pack('C8')
+  MAGIC_BYTES = [RW2_MAGIC_BYTES, RAW_RWL_MAGIC_BYTES]
+  BORDER_TAG_IDS = {
+    top: 4,
+    left: 5,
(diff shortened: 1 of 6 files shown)


spec/active_storage/rails_app_spec.rb
```diff
@@ -6,10 +6,11 @@ def skip_reason
   elsif RUBY_VERSION.to_f < 2.5
     'Skipping because Rails testing script use Rails 6, who does not support Ruby bellow 2.5'
   else
-    false
+    'Skipping because this test randomly started failing for every version - mismatching default gem versions.'
```
Kevin-McGonigle, line 9: There is some _seriously_ weird stuff happening here. From what I can see, the bundler of the ruby process that is spun up inside the test tries to install versions of the default gems that do not match those in the outer process'. No idea why this has started happening - we never used to see these gems being installed before now. We should create an issue for this and address it ASAP so we can reintroduce this test for regression testing. (Note: This test _does_ pass locally on its _second_ run).

lib/parsers/rw2_parser.rb
```diff
@@ -0,0 +1,61 @@
...
+      intrinsics: { exif: exif },
+      content_type: PANASONIC_RAW_MIMETYPE,
+    )
+  rescue EXIFR::MalformedTIFF
+    nil
+  end
+
+  private
+
+  PANASONIC_RAW_MIMETYPE = 'image/x-panasonic-raw'
```
linkyndy, line 41: These constants are **not** private. Don't think they need to be, thus I'd move them to the usual spot (under the class declaration). Otherwise, there's an _opinionated_ modifier for that.

lib/parsers/rw2_parser.rb
```diff
@@ -0,0 +1,61 @@
...
+    bottom: 6,
+    right: 7
+  }
+
+  def matches_rw2_definition?
+    MAGIC_BYTES.include?(read_bytes(8))
+  end
+
+  def sensor_border(exif, border)
+    exif[0].raw_fields[BORDER_TAG_IDS[border]][0]
```
linkyndy, line 57: Just double checking this won't throw any IndexError.

spec/active_storage/rails_app_spec.rb
linkyndy: Did it start failing with this PR? 🤔

spec/parsers/rw2_parser_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'spec_helper'
+
+describe FormatParser::RW2Parser do
+  context 'when likely_match? is called' do
+    %w[rw2 RW2 raw RAW rwl RWL].each do |extension|
+      context "with a file with a .#{extension} extension" do
+        it "should return true" do
```
linkyndy, line 7:
```suggestion
        it "returns true" do
```

as per [better specs](https://www.betterspecs.org/).

spec/parsers/rw2_parser_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+          expect(subject.likely_match?("foo.#{extension}")).to be_truthy
+        end
+      end
+    end
+
+    ['', 'foo', 'rw2', 'foorw2', 'foo.rw', 'foo.rw2.bar'].each do |filename|
+      context "with a file named #{filename}" do
+        it 'should return false' do
+          expect(subject.likely_match?(filename)).to be_falsey
```
linkyndy, line 16: Just bringing this up: `be_falsey` is not really `be(false)`; the latter implies it, but it's not going both ways.

lib/parsers/rw2_parser.rb
```diff
@@ -19,8 +30,13 @@ def call(io)
 
     # RW2 doesn't use the standard EXIF width and height tags (🤷🏻). We can compute them from the sensor
     # top/bottom/left/right border tags. See https://exiftool.org/TagNames/PanasonicRaw.html for more.
-    w = sensor_border(exif, :right) - sensor_border(exif, :left)
-    h = sensor_border(exif, :bottom) - sensor_border(exif, :top)
+    left_sensor_border = sensor_border(exif, :left)
+    right_sensor_border = sensor_border(exif, :right)
+    w = left_sensor_border && right_sensor_border && right_sensor_border - left_sensor_border
```
linkyndy, line 35: I find this a bit confusing; should it return a boolean? Is it `nil` or `false` when the conditions are not met? `&&` and `||` are traditionally used when after a boolean value, so I would write these in a more _expected_ way.

linkyndy (approved):

rcpalacio (approved):
I'm assuming we have a ticket for bringing back the skipped the tests

Kevin-McGonigle:
[CORE-262](https://wetransfer.atlassian.net/browse/CORE-262)

PR #227, opened 2023-01-31
[CORE-257] Fix MP4 Misidentification

Kevin-McGonigle:
This PR addresses issues with the previous implementation of MOV/MP4 parsing, including:
- Misidentification of MP4 files as MOV files.
- Incorrect dimensions and other metadata for files that had multiple tracks and/or transformation matrices.

Also taking the opportunity to adopt the ISOBMFF decoder that was introduced as part of the CR3 support work.


rcpalacio (approved):
Looks great! 
Should we have any mention to an overview or spec of the iso base media file format and how other formats make use of it?

lib/parsers/iso_base_media_file_format/decoder.rb
```diff
@@ -1,30 +1,42 @@
...
-      class Atom < Struct.new(:type, :position, :size, :fields, :children)
+      class Box < Struct.new(:type, :position, :size, :fields, :children)
         def initialize(type, position, size, fields = nil, children = nil)
           super
           self.fields ||= {}
           self.children ||= []
         end
 
+        def [](index)
+          if index.is_a?(Symbol)
```
linkyndy, line 27: I find this a bit confusing. Wouldn't expect that addressing this class by symbol or string would look in different places for that particular element 🤔

lib/parsers/iso_base_media_file_format/decoder.rb
```diff
@@ -417,137 +441,137 @@ def empty(size)
...
-      #     duration: version == 1 ? read_int_64 : read_int_32,
-      #     rate: read_fixed_point_32,
-      #     volume: read_fixed_point_16,
-      #     matrix: skip_bytes(10) { read_matrix },
-      #     next_trak_id: skip_bytes(24) { read_int_32 },
-      #   })
-      #   [fields, nil]
-      # end
+      # Parse a movie header box.
+      def mvhd(_)
```
linkyndy, line 551: Will this argument be used in the future?

lib/parsers/iso_base_media_file_format/decoder.rb
```diff
@@ -1017,22 +1041,16 @@ def uuid(size)
...
       # See https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap4/qtff4.html#//apple_ref/doc/uid/TP40000939-CH206-18737.
       def read_matrix
-        9.times.map do |i|
-          if i % 3 == 2
-            read_fixed_point_32_2_30
-          else
-            read_fixed_point_32
-          end
-        end
+        Matrix.build(3) { |_, c| read_fixed_point(fractional_digits: c % 3 == 2 ? 30 : 16, signed: true) }
```
linkyndy, line 1044: Curious, does `matrix` significantly improve this logic, as opposed to the more rudimentary approach we currently have?

lib/parsers/iso_base_media_file_format/utils.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'matrix'
+
+module FormatParser::ISOBaseMediaFileFormat::Utils
+  def codecs(box_tree)
+    video_trak_boxes(box_tree).map do |trak_box|
```
linkyndy, line 7:
```suggestion
    video_trak_boxes(box_tree).flat_map do |trak_box|
```

lib/parsers/iso_base_media_file_format/utils.rb
```diff
@@ -0,0 +1,85 @@
...
+      stsd_box = trak_box['mdia']['minf']['stbl']['stsd']
```
linkyndy, line 8: Are we sure about the existence of all these keys, or shall we rather `dig` and handle `nil`s?

lib/parsers/iso_base_media_file_format/decoder.rb
Kevin-McGonigle: I thought the pros out-weighed the cons in favour of this approach over separate, dedicated methods - I think it gives more clarity when trying to access deeply nested boxes, e.g. `trak_box['mdia']['minf']['stbl']['stsd']` vs. `trak_box.get_child('mdia').get_child('minf).get_child('stbl').get_child('stsd')` vs `trak_box.get_descendent(['mdia', 'minf', 'stbl', 'stsd']`.

I don't feel especially strongly about it though, so if you do then I'm happy to change it up. ✌🏻

lib/parsers/iso_base_media_file_format/utils.rb
Kevin-McGonigle: All of these boxes are specified as mandatory in the ISOBMFF spec, so if they don't exist then that indicates that the file is invalid or that we have a bug elsewhere in the code - both of which warrant an error being thrown imo.

README.md
```diff
@@ -195,16 +199,15 @@ Unless specified otherwise in this section the fixture files are MIT licensed an
 ### M3U
 - The M3U fixture files were created by one of the project maintainers
 
-### M4A
```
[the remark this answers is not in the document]
Kevin-McGonigle: I lumped it in with the MP4 fixtures.

lib/parsers/iso_base_media_file_format/decoder.rb
Kevin-McGonigle: _Significantly improve?_ Probably not, but feel like it represents the intentions of the matrix fields better than a flat 9-element array, and the value is then immediately ready for use in any calculations.

lib/parsers/iso_base_media_file_format/decoder.rb
linkyndy: I'm fine either way. Just pointed out that it looks confusing at first sight; I trust you had a better overview of the usage of this 🙌

lib/parsers/iso_base_media_file_format/decoder.rb
linkyndy: Got it 👌

linkyndy (approved):
Great effort, Kevin! ❤️

PR #229, opened 2023-02-27
[CORE-452] ISOBMFF Resiliency

Kevin-McGonigle:
* Handle cases where mandatory boxes or fields are missing.

* Simplify frame-rate calculation.

* Refactor.

* Lots of tests.


lib/parsers/iso_base_media_file_format/utils.rb
```diff
@@ -2,82 +2,86 @@
...
+      def frame_rate(box_tree)
+        video_trak_boxes(box_tree).each do |trak_box|
+          mdhd_box = trak_box.first_descendent_by_path(%w[mdia mdhd])
+          stts_box = trak_box.first_descendent_by_path(%w[mdia minf stbl stts])
 
-      timescale = mdhd_box[:timescale].to_f
-      stts_entries = stts_box[:entries]
+          next unless mdhd_box && stts_box
 
-      frame_rate_for_entry = ->(stts_entry) { (timescale / stts_entry[:sample_delta]).truncate(2) }
```
Kevin-McGonigle, line 56: I think ditching this plus [returning from the loop when we first find components needed to calculate a frame rate](https://github.com/WeTransfer/format_parser/pull/229/files#diff-2f106d6b2af784f07e591e9480fd4cf955773d879d2a24dc211fe01768c9f4fcR58) will save us from the `SystemStackError` we were seeing in some instances. We won't needlessly iterate through all of the time-to-sample boxes as we were before.

lib/parsers/iso_base_media_file_format/box.rb
```diff
@@ -0,0 +1,80 @@
+module FormatParser
+  module ISOBaseMediaFileFormat
+    class Box < Struct.new(:type, :position, :size, :fields, :children)
+      def initialize(type, position, size, fields = nil, children = nil)
+        super
+        self.fields ||= {}
```
linkyndy, line 6: You don't need an `initialize` for this; you can simply define a:

```
def fields
  @fields ||= {}
end
```

Or, seeing that you need so much more customisation on top of what Struct gives, you might be better off not using it at all (you're only saving a few assignments when initialising the class).

lib/parsers/iso_base_media_file_format/box.rb
```diff
@@ -0,0 +1,80 @@
...
+      def all_children(*types)
+        children.select { |child| types.include?(child.type) }
+      end
+
+      # Returns true if there are one or more children with the given type.
+      #
+      # @param [String] type The box type to search for.
+      # @return [Boolean]
+      def child?(type)
+        !(children.select { |child| child.type == type }).empty?
```
linkyndy, line 23:
```suggestion
        children.any? { |child| child.type == type }
```

lib/parsers/iso_base_media_file_format/box.rb
```diff
@@ -0,0 +1,80 @@
...
+      def first_child(*types)
+        children.find { |child| types.include?(child.type) }
+      end
+
+      # Find and return all descendents of a given type.
+      #
+      # @param [Array<String>] types The box type(s) to search for.
+      # @return [Array<Box>]
+      def all_descendents(*types)
+        children.map do |child|
```
linkyndy, line 39: You can use `flat_map` to make it a _bit_ faster and shorter.

lib/parsers/iso_base_media_file_format/box.rb
```diff
@@ -0,0 +1,80 @@
...
+          descendents = child.all_descendents(*types)
+          types.include?(child.type) ? [child] + descendents : descendents
+        end.flatten
+      end
+
+      # Find and return all descendents that exists at the given path.
```
linkyndy, line 45:
```suggestion
      # Find and return all descendents that exist at the given path.
```

lib/parsers/iso_base_media_file_format/utils.rb
```diff
@@ -2,82 +2,86 @@
...
+          timescale = mdhd_box.fields[:timescale]&.to_f
+          sample_delta = stts_box.fields[:entries]&.[](0)&.[](:sample_delta)
```
linkyndy, line 54: I'm wondering, wouldn't something like this work?

```suggestion
          sample_delta = stts_box.fields.dig(:entries, 0, :sample_delta)
```

spec/parsers/iso_base_media_file_format/box_spec.rb
```diff
@@ -0,0 +1,399 @@
+require 'spec_helper'
+
+describe FormatParser::ISOBaseMediaFileFormat::Box do
```
linkyndy, line 3: <3

spec/parsers/iso_base_media_file_format/utils_spec.rb
```diff
@@ -0,0 +1,632 @@
+require 'matrix'
+require 'spec_helper'
+
+describe FormatParser::ISOBaseMediaFileFormat::Utils do
+  include FormatParser::ISOBaseMediaFileFormat::Utils
+  Box = FormatParser::ISOBaseMediaFileFormat::Box
```
linkyndy, line 6: If you're testing `Box`, then I'd `describe` it and use it directly as `described_class` and/or `subject` (if it makes sense), rather than manually making it available under a different module's `describe`. I also wouldn't mix in a tested module into a testing block.

lib/parsers/iso_base_media_file_format/utils.rb
Kevin-McGonigle: Wasn't aware of dig - awesome!

spec/parsers/iso_base_media_file_format/utils_spec.rb
Kevin-McGonigle: Not testing `Box`, testing the methods in the `Utils` module and each of those methods take a `box_tree` as an argument - so I need to use the `Box` class to set up the appropriate tests.

Since we're testing the methods in the module, the only options I found to test them were to:
1. `include` the `Utils` module and test the methods directly, as I've done here.
2. Set up a dummy class that includes the module and use that, like:
   ```ruby
   let(:dummy) { (Class.new { include FormatParser::ISOBaseMediaFileFormat::Utils }).new }
   ```

I opted for 1 over 2 because 2 seemed needlessly complex to me and made the tests a little harder to understand, since every method would have to be called as `dummy.method(box_tree)` instead of just calling the method under test directly.

lib/parsers/iso_base_media_file_format/box.rb
Kevin-McGonigle: Hmm, removing the `initialize` def and adding getter methods for `fields` and `children` as you suggested causes several test failures - so it's changing the behaviour and I'm not really sure how.

We could ditch Struct, but it gives us more than just a constructor. For example, we get a more useful string representation a `to_h` method and equality operators, all of which are useful for debugging. Also, it gives us a `dig` method which works well with your suggestion down below to avoid chaining `&.` operators - so I'd rather keep it as-is.

lib/parsers/iso_base_media_file_format/box.rb
linkyndy: There were a few other occurrences of this 😇

spec/parsers/iso_base_media_file_format/utils_spec.rb
linkyndy: If you're testing methods inside a module, you can do that directly, without any intermediary, and provide `Box` as an argument, defined via an overrideable `let`:

```
describe FormatParser::ISOBaseMediaFileFormat::Utils do
  let(:box) { Box.new(...) }

  describe '.method' do
    let(:box) { Box.new(...) }
  end
end
```

I would go for your second approach if I had to test a class/module which had the utils **included**, for whatever reason.

lib/parsers/iso_base_media_file_format/box.rb
linkyndy: Probably you need `self.fields ||= {}` instead of `@fields ||= {}`. Also, if you're not using any args in `initialize`, you can just `initialize(*args)` them.

I see, if there are multiple benefits of the Struct, let's keep it.

spec/parsers/iso_base_media_file_format/utils_spec.rb
Kevin-McGonigle: I think we're getting wires crossed here. Line 6 here:
```ruby
Box = FormatParser::ISOBaseMediaFileFormat::Box
```
is purely to avoid having to fully qualify the namespace when creating each of the 200+ `Box` instances required to set up the appropriate `box_tree` arguments (which I achieve using `let`) throughout, which greatly improves readability imo.

For example, without that line, this:
```ruby
let(:box_tree) do
  [
    Box.new('moov', 0, 0, nil, [
      Box.new('trak', 0, 0),
      Box.new('trak', 0, 0, nil, [
        Box.new('hdlr', 0, 0, { handler_type: 'soun' }),
      ]),
      Box.new('trak', 0, 0, nil, [
        Box.new('hdlr', 0, 0, { component_type: 'mhlr', component_subtype: 'soun' }),
      ])
    ])
  ]
end
```

instead of:

```ruby
let(:box_tree) do
  [
    FormatParser::ISOBaseMediaFileFormat::Box.new('moov', 0, 0, nil, [
      FormatParser::ISOBaseMediaFileFormat::Box.new('trak', 0, 0),
      FormatParser::ISOBaseMediaFileFormat::Box.new('trak', 0, 0, nil, [
        FormatParser::ISOBaseMediaFileFormat::Box.new('hdlr', 0, 0, { handler_type: 'soun' }),
      ]),
      FormatParser::ISOBaseMediaFileFormat::Box.new('trak', 0, 0, nil, [
        FormatParser::ISOBaseMediaFileFormat::Box.new('hdlr', 0, 0, { component_type: 'mhlr', component_subtype: 'soun' }),
      ])
    ])
  ]
end
```

linkyndy: Right, fair enough, that makes sense 😊

lib/format_parser/version.rb
```diff
@@ -1,3 +1,3 @@
 module FormatParser
-  VERSION = '2.4.1'
```
linkyndy, line 2: Just be aware that there is another PR in the works, which _might_ be merged before this one ✌️

lib/parsers/iso_base_media_file_format/utils.rb
```diff
@@ -62,9 +62,9 @@ def frame_rate(box_tree)
       end
 
       def video_codecs(box_tree)
-        video_trak_boxes(box_tree).map do |trak_box|
-          trak_box.all_descendents_by_path(%w[mdia minf stbl stsd]).map { |stsd_box| stsd_box.children.map(&:type) }
-        end.flatten.compact.uniq
+        video_trak_boxes(box_tree).flat_map do |trak_box|
+          trak_box.all_descendents_by_path(%w[mdia minf stbl stsd]).flat_map { |stsd_box| stsd_box.children.map(&:type) }
```
linkyndy, line 66: Do we need this nested `flat_map` as well? Looking at the diff, it's different than before.

linkyndy (approved):

lib/parsers/iso_base_media_file_format/utils.rb
Kevin-McGonigle: We either need a nested `flat_map` _or_ we need to revery to the `map` and `flatten` on the outer block, because `flat_map` != `map` & `flatten`.

`flat_map` will only flatten to a depth of 1 and will not totally flatten arrays of arrays.

PR #232, opened 2023-03-27
[NO JIRA] Prevent 0 Size Infinite Loop

Kevin-McGonigle:
When an ISO Base Media File Format box has a size of 0, the box should be assumed to extend to the end of the file. We do not currently account for this scenario, leading to infite loops where the same box (perceived to have no size) is "parsed" over and over again (and a bunch of buffer position correction warnings). This PR fixes this oversight.

### DIFF
--- a/spec/parsers/iso_base_media_file_format/decoder_spec.rb
+++ b/spec/parsers/iso_base_media_file_format/decoder_spec.rb
@@ -48,12 +48,29 @@ describe FormatParser::ISOBaseMediaFileFormat::Decoder do
       end
       let(:result) { subject.build_box_tree(0xFF, io) }
 
-      it('parses successfully') { expect(result.length).to eq(1) }
-      it('parses the correct type') { expect(result[0].type).to eq('foo ') }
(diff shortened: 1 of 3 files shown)


spec/parsers/iso_base_media_file_format/decoder_spec.rb
```diff
@@ -113,5 +113,30 @@
...
+        # foo
+        # moov
+        # |-> bar
+        # |-> baz
+        input = [0x8].pack('N') + 'foo ' + [0x0].pack('N') + 'moov' + [0x8].pack('N') + 'bar ' + [0x8].pack('N') + 'baz '
+        StringIO.new(input)
+      end
+      let(:result) { subject.build_box_tree(0xFF, io) }
+
+      it('reads the rest of the file') do
```
linkyndy, line 128: Not a biggie, but I've never seen parens with `it`s before; it just looks odd 😀

linkyndy (approved):
🙏

RubenWolff-wt (approved):
lgtm

PR #233, opened 2023-03-28
Disable `udta` Parsing

Kevin-McGonigle:
The User Data (`udta`) ISOBMFF box is currently treated as a standard container box, expecting other ISOBMFF boxes within. However, their contents don't always necesarily conform to the standard, resulting in some incorrect parsing, corrective buffer position updates and warning logs. This PR disables `udta` box parsing, which will skip parsing their contents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.4.5
+* Disable `udta` ISOBMFF box parsing, since their contents are not guaranteed to be consistent with the spec.
+
 ## 2.4.4
 * Prevent infinite loops when parsing ISOBMFF boxes with size = 0 (meaning that the box extends to the end of the file).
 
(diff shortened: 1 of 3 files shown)


RubenWolff-wt (approved):
lgtm

I have no idea what exactly that change does but it matches your description :)

Kevin-McGonigle:
> lgtm
> 
> I have no idea what exactly that change does but it matches your description :)

If you're interested, head to peek-worker logs and you'll see a bunch that look like:
```
Unexpected IO position after parsing udta box at position X. Box size: Y. Expected position: X + Y. Actual position: Z.
```

Basically, we try to treat the contents of the `udta` box as if it confomed to the standard, but it often doesn't, so our parser freaks out and ends up in a weird state. We have a fail-safe built in that resets the buffer to the correct position and generates that log warning when something like this happens - but it's happening pretty often for `udta` boxes and polluting the logs a bit. So this PR basically stops treating the `udta` box as a container, which was prompting the parser to attempt to parse its contents, and now treats it as an unknown box, which will now just note the existence of the box without diving inside it.

linkyndy (approved):
Appreciate the nice explanation as well 🙌

PR #234, opened 2023-05-10
add avc1 and xavc formats to mp4 parser

RubenWolff-wt:
The new sub formats (brands?) showed up in some of our use cases.

For now I have been unable to find test files with a license of the correct sub format of mp4 to be able to test these and validate my change works.

I added webrick as a dev dependency due to this causing errors in the [ruby 3.1 build](https://github.com/WeTransfer/format_parser/actions/runs/4936873971/jobs/8824943012). I found a related issue that suggests [webrick is no longer bundled in ruby 3.x and fixing it by adding it as a dependency](https://github.com/jekyll/jekyll/issues/8523). As we only use it in a test file I only added it as a dev dependency to keep the changes minimal. I'm not sure how this built successfully previously 🤔 

**UPDATE** Both myself and @rcpalacio have been unable to find any licensed test files. As we have some open issues in portal that this may fix I have discussed with Vanja that we would be ok to add this fix without specific test files and add these later if we find any.

### DIFF
--- a/format_parser.gemspec
+++ b/format_parser.gemspec
@@ -42,7 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'webmock'
+  spec.add_development_dependency 'webrick'
   spec.add_development_dependency 'wetransfer_style', '1.0.0'
   spec.add_development_dependency 'yard'
-  spec.add_development_dependency 'webrick'
 end
--- a/format_parser.gemspec
+++ b/format_parser.gemspec
@@ -44,4 +44,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'webmock'
   spec.add_development_dependency 'wetransfer_style', '1.0.0'
   spec.add_development_dependency 'yard'
+  spec.add_development_dependency 'webrick'
 end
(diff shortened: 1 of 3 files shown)


format_parser.gemspec
```diff
@@ -44,4 +44,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'webmock'
   spec.add_development_dependency 'wetransfer_style', '1.0.0'
   spec.add_development_dependency 'yard'
+  spec.add_development_dependency 'webrick'
```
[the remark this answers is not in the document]
RubenWolff-wt: addressed

Kevin-McGonigle (approved):

rcpalacio (approved):
👍

format_parser.gemspec
```diff
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'webmock'
+  spec.add_development_dependency 'webrick'
```
linkyndy, line 45: afaik, webrick was extracted as a separate Ruby gem and thus that might be why it was causing issues with the latest Ruby versions.

linkyndy (approved):
Would be great if we could find some examples to test these new changes against 👍

format_parser.gemspec
RubenWolff-wt: Yeah that's what I found out as well. Really odd this was not erroring earlier as I don't think this has changed that recently 🤔

PR #235, opened 2023-05-11
update gem version

RubenWolff-wt:


linkyndy (approved):
I'd consider this a minor change, as patches are usually fixing bugs or vulnerabilities.

linkyndy (approved):

PR #237, opened 2023-05-30
Improving I/O consumption for TIFF-based formats

rcpalacio:
All parsers for TIFF-based formats were loading embedded thumbnails alongside exif data, which was blowing the IO budget for large images.

This upgrades the dependency EXIFR that now allows us to skip this step.

(the actual work was done [here](https://github.com/remvee/exifr/pull/70))

### DIFF
--- a/lib/parsers/exif_parser.rb
+++ b/lib/parsers/exif_parser.rb
@@ -175,7 +175,7 @@ module FormatParser::EXIFParser
   def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
     Measurometer.instrument('format_parser.exif_parser.exif_from_tiff_io') do
       extended_io = IOExt.new(constrained_io)
-      exif_raw_data = EXIFR::TIFF.new(extended_io)
+      exif_raw_data = EXIFR::TIFF.new(extended_io, load_thumbnails: false)
 
       return unless exif_raw_data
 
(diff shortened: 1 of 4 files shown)


enricoribelli (approved):

lib/parsers/exif_parser.rb
```diff
@@ -175,7 +175,7 @@ def method_missing(*a)
   def exif_from_tiff_io(constrained_io, should_include_sub_ifds = false)
     Measurometer.instrument('format_parser.exif_parser.exif_from_tiff_io') do
       extended_io = IOExt.new(constrained_io)
-      exif_raw_data = EXIFR::TIFF.new(extended_io)
+      exif_raw_data = EXIFR::TIFF.new(extended_io, load_thumbnails: false)
```
linkyndy, line 178: One question: could it be useful to provide this as a config for the library? I can imagine some might actually need those thumbnails.

linkyndy (approved):
Looks great, thanks for syncing with the other library 💪 

Just a small remark; feel free to move forward if you disagree with it.

lib/parsers/exif_parser.rb
rcpalacio: You mean via `intrinsics`? That is a good point. I'll change this. 
On this: are the intrinssics something we want to maintaining stable across multiple versions?

linkyndy: Not necessarily, just a way to configure this when using the library.

On the intrinsics, I think it would be nice, though I wouldn't consider it as a priority.

PR #239, opened 2023-06-26
JSON format support

rcpalacio:
This adds a new parser to support JSON files.

The new parser tries to validate the JSON syntax in a file without actually parsing and loading a JSON object in memory for both performance and security reasons.


lib/parsers/json_parser/validator.rb
```diff
@@ -0,0 +1,318 @@
...
+  end
+
+  def stats(node_type)
+    @execution_stats[node_type]
+  end
+
+  private
+
+  def setup_transitions
+    when_its :awaiting_root_node, ->(c) {
```
linkyndy, line 74: I'd use `do...end` for multiline blocks.

lib/parsers/json_parser/validator.rb
```diff
@@ -0,0 +1,318 @@
...
+      read_whitespace(c) or
+        start_object(c) or
```
linkyndy, line 76: Curious whether there's any reason for the extra indentation here 🤔

lib/parsers/json_parser/validator.rb
```diff
@@ -0,0 +1,318 @@
...
+    @all_parsers[state] = act
+  end
+
+  def parse_char(c)
+    next_step = @all_parsers[@current_state]
+    accepted = next_step.call(c)
+    reject_char(c) unless accepted
+  end
+
+  def read_whitespace(c)
```
linkyndy, line 149: I would go with `?` methods here, given they return a bool. Looks more Ruby-like 😊

lib/parsers/json_parser/validator.rb
```diff
@@ -0,0 +1,318 @@
...
+    begin_node(:literal)
+    @current_state = :reading_literal
+    @current_literal_size = 1
+    true
+  end
+
+  def close_literal(c)
+    raise JSONParserError, "Literal to large at #{@pos}" if @current_literal_size > MAX_LITERAL_SIZE
+
+    if whitespace?(c) or c == "," or c == "]" or c == "}"
```
linkyndy, line 275: Usually `||` is preferred, unless there a specific reason to use `or`. I'd be inclined to make the same suggestion for the `when_its`s above (is it like just for readability or it's an actual precedence requirement?)

lib/parsers/json_parser/validator.rb
```diff
@@ -0,0 +1,318 @@
...
+  end
+
+  def control_char?(c)
+    # control characters: (U+0000 through U+001F)
+    utf8_code = c.unpack('U*')[0]
+    utf8_code <= 31
+  end
+
+  def valid_literal_char?(c)
+    LITERALS_CHAR_TEMPLATE === c
```
linkyndy, line 316: I think using `c.match?(regex)` is a bit more readable in this case 😊

lib/utf8_reader.rb
```diff
@@ -0,0 +1,66 @@
...
+    @index = 0
+    @eof = true if @chunk.nil? or @chunk.length < READ_CHUNK_SIZE
+  end
+
+  def assess_char_length(first_byte)
+    # 0_______ (1 byte)
+    # 110_____ (2 bytes) 192
+    # 1110____ (3 bytes) 224
+    # 11110___ (4 bytes) 240
+    return 4 if first_byte >= 240
```
linkyndy, line 61: A `case` would work nicely in this case 😊

enricoribelli:
Could you also update the README and up the version in `version.rb` ?

lib/parsers/json_parser.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  JSON_MIME_TYPE = 'application/json'
+
+  def likely_match?(filename)
+    filename =~ /\.json$/i
+  end
+
+  def call(io)
+    io = FormatParser::IOConstraint.new(io)
+    io.seek(0)
```
enricoribelli, line 13: I believe the parser starts from the beginning. Is the seek/positioning needed?

lib/parsers/json_parser/validator.rb
rcpalacio: Rubocop requires the additional indentation for expressions spawning multiple lines.

lib/parsers/json_parser/validator.rb
rcpalacio: @linkyndy 
Do you mean rename it to `read_whitespace?`  or replace this with the existing `whitespace?` method?

The reason having `?` seemed off to me was that these methods usually have side effects, which I don't expect a `?` method to have. (For instance,  `read_colon` updates @current_state.
These methods are more about performing an action and reporting success than actually answering a question.

I can see `read_whitespace?`, `read_valid_string_char` or `close_array` working, for example, but I wouldn't go for replacing with `whitespace?` altogether.

Wdyt?

lib/parsers/json_parser/validator.rb
rcpalacio: It was a deliberate decision to use `or` with `when_its` for the sake of looking like english statements. does that one look weird?

here, I'll replace this with an array `include?` instead, actually... but I agree `||` is better than `or` for the use case.

lib/utf8_reader.rb
```diff
@@ -0,0 +1,66 @@
+##
+# This class Reads individual characters from files using UTF-8 encoding
+# This deals with two main concerns:
+#   - Variable byte length of characters
+#   - Reducing the number of read operations by loading bytes in chunks
+
+class FormatParser::UTF8Reader
```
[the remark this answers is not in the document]
rcpalacio: you mean the `safe_read` method for dealing with errors raised in EOF or the `IOConstraint` class?

as for `IOConstraint`, it is in use by default, as that's what a parser hands over to this validator.
`safe_read` can be used here if you think it's necessary, as long as we rescue those to new possible errors down the line.

Any thoughts?

lib/parsers/json_parser/validator.rb
```diff
@@ -0,0 +1,318 @@
...
+    @current_node = nil # :object, :array, :string, :literal
+    @parent_nodes = []
+    @current_state = :awaiting_root_node
+    @escape_next = false
+    @current_literal_size = 0
+    @pos = 0
+
+    @all_parsers = {}
+
+    @execution_stats = {
```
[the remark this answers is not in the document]
rcpalacio: That is a good point. IDK if it would be more useful or misleading, as the ISO standard is not implemented by the letter. We end up with string counts that also regards attribute names; literals are also grouping together a bunch of stuff.
Do you see actual benefit?

lib/parsers/json_parser/validator.rb
linkyndy: Interesting!

lib/parsers/json_parser/validator.rb
linkyndy: If we're sure the resulting precedence works in our favour, `or`s are just fine 😊

lib/utf8_reader.rb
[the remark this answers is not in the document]
linkyndy: The former. If we already have the logic to read N bytes from various sources and deal with potential errors and such, I was curious why also (re)implementing this here. Probably I'm missing the larger context of how `safe_read` works 🤔

lib/parsers/json_parser/validator.rb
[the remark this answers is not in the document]
linkyndy: Then it's probably best if we don't do that, agree.

lib/utf8_reader.rb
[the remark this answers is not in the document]
rcpalacio: the point of this UTF8 reader is: we don't actually know how many bytes we want to read, since chars can have different lenghts - from 1 to 4. This reader abstracts that away and gives us back chars one by one, no matter their size.

internally it could use safe_read, but that would only deal with `EOF`, which is not a problem at all here. 

```
def safe_read(io, n)
    raise ArgumentError, 'Unbounded reads are not supported' if n.nil?
    buf = io.read(n)

    raise InvalidRead, "We wanted to read #{n} bytes from the IO, but the IO is at EOF" unless buf
    raise InvalidRead, "We wanted to read #{n} bytes from the IO, but we got #{buf.bytesize} instead" if buf.bytesize != n

    buf
  end

```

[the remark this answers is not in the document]
linkyndy: Right, that makes more sense 👍

lib/utf8_reader.rb
linkyndy: Right. I was having in mind something like [this](https://stackoverflow.com/a/22700292/433940). It's arguable whether it's more readable or not.

README.md
```diff
@@ -25,6 +25,7 @@ and [dimensions,](https://github.com/sstephenson/dimensions) borrowing from them
 * GIF
 * HEIC
 * HEIF
+* JSON
```
linkyndy, line 28: Just the alphabetical order 🙈

lib/parsers/json_parser/validator.rb
```diff
@@ -71,69 +72,69 @@ def stats(node_type)
   private
 
   def setup_transitions
-    when_its :awaiting_root_node, ->(c) {
+    when_its :awaiting_root_node, ->(c) do
       read_whitespace(c) or
         start_object(c) or
         start_array(c)
-    }
+      end
```
linkyndy, line 79: Isn't this supposed to align with `when_its`? 🤔

lib/utf8_reader.rb
```diff
@@ -58,9 +58,11 @@ def assess_char_length(first_byte)
     # 110_____ (2 bytes) 192
     # 1110____ (3 bytes) 224
     # 11110___ (4 bytes) 240
-    return 4 if first_byte >= 240
-    return 3 if first_byte >= 224
-    return 2 if first_byte >= 192
-    1
+    case first_byte
+    when 240..Float::INFINITY then 4
```
linkyndy, line 62:
```suggestion
    when 240.. then 4
```

this should do 😊

linkyndy (commented):
👌

README.md
rcpalacio: good catch!

linkyndy (approved):
🔥

PR #240, opened 2023-08-22
Fixed bug with WAV file wrongly parsed as MP3

CostinTanasoiu:
This happened when parsing files without passing a filename hint, hence relying only on the file bytes. The MP3 format is loose and the MP3 parser is prone to false positives, but hopefully the following fixes will prevent this from happening for any of the file types for which we have registered parsers.

- Changed MP3 priority from 99 to 101. This ensures that MP3 parser is the last one to be used when we're iterating through parsers.
- Added a preventive check against WAV headers in the MP3 parser
- Added a new test verifying that we correctly parse files over HTTP even without a filename hint
- Added a test verifying that MP3 is always the last parser used when not providing a filename hint
- Set up a new convention that invalid fixtures contain the "invalid" word in their filename and added tests for this
- Updated some fixture names


spec/format_parser_spec.rb
```diff
@@ -189,12 +189,26 @@
...
+        'FormatParser::WAVParser',
+        'FormatParser::MP3Parser'
       ])
     end
+
+    it 'ensures that MP3 parser is the last one among all' do
+      parsers = FormatParser.parsers_for(
+        [:audio, :image, :document, :text, :video, :archive],
+        [
+          :aac, :aiff, :arw, :bmp, :cr2, :cr3, :dpx, :fdx, :flac, :gif, :heif, :heic,
```
rcpalacio, line 203: _(Same comment as that from `spec/remote_fetching_spec.rb`)_ 

> It feels more future-proof to get this programatically, maybe exposing FormatParser's `@parsers_per_nature` somehow? or exposing all parsers as an array in proper execution order?

lib/parsers/mp3_parser.rb
```diff
@@ -315,5 +320,5 @@ def with_id3tag_local_configs
     end
   end
 
-  FormatParser.register_parser new, natures: :audio, formats: :mp3, priority: 99
+  FormatParser.register_parser new, natures: :audio, formats: :mp3, priority: 101
```
linkyndy, line 323: Just for the record: which other parsers may be impacted by the change of priority here? As far as I know, the MP3 parser was at the end of the line at some point in the past, were there other parsers added after it? And if so, would the ordering mess up them?

spec/format_parser_spec.rb
linkyndy: Should we expose an internal state just because it would simplify some tests? 🤔

spec/remote_fetching_spec.rb
```diff
@@ -104,6 +104,57 @@
...
+          if file_type_dir == 'HEIF'
+            expected_format = File.extname(file_name).delete('.').downcase.to_sym
+          end
+
+          it "parses #{file_type_dir} file: #{file_name}" do
+            url = "http://localhost:9399/#{file_type_dir}/#{file_name}?some_param=test".gsub(' ', '%20')
+            file_information = FormatParser.parse_http(url)
+            expect(file_information).not_to be_nil
+            expect(file_information.nature).to eq(nature)
+            expect(file_information.format == expected_format).to be_truthy
```
linkyndy, line 141:
```suggestion
            expect(file_information.format).to eq(expected_format)
```

I find this a bit more "expected", also it's clearer from the type point of view.

spec/remote_fetching_spec.rb
```diff
@@ -104,6 +104,57 @@
...
+          end
+        end
+      end
+    }
+
+    # Dir.glob(fixtures_dir + "/HEIF/*.*").each do |file_path|
```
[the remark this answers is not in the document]
CostinTanasoiu: Removed.

spec/remote_fetching_spec.rb
```diff
@@ -104,6 +104,57 @@
...
+      document: ['PDF'],
+      audio: ['AAC', 'AIFF', 'FLAC', 'MP3', 'WAV'],
+      video: ['MOV', 'MP4'],
+      image: ['ARW', 'CR2', 'CR3', 'GIF', 'JPG', 'NEF', 'PNG', 'PSD', 'RW2', 'TIF', 'WEBP']
+    }
+    nature_fixture_dirs.each { |nature, dirs|
+      dirs.each do |file_type_dir|
+        Dir.glob(fixtures_dir + "/#{file_type_dir}/*.*").each do |file_path|
+          file_name = File.basename(file_path)
+          next if file_name.include? "invalid"
```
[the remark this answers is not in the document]
CostinTanasoiu: I've added two new tests in `format_parser_spec.rb`:
- "fixtures with 'invalid' in the filename should fail to parse"
- "fixtures without 'invalid' in the filename should be parsed successfully"

lib/parsers/mp3_parser.rb
CostinTanasoiu: The order for the remaining parsers doesn't change, aside from pushing MP3 to the end.

Most parsers did not have a set priority, so they defaulted to 99 on registration – the same as MP3 which had priority 99 hardcoded. Here is what the state of priorities looked like before my change:
image
all those marked with red are defaulting to 99...

spec/remote_fetching_spec.rb
```diff
@@ -104,6 +104,57 @@
...
+        remote_url = 'http://localhost:9399/WAV/' + filename
+        file_information = FormatParser.parse_http(remote_url)
+        expect(file_information).not_to be_nil
+        expect(file_information.format).to eq(:wav)
+      end
+    end
+  end
+
+  describe "correctly parses files over HTTP without filename hint" do
+    nature_fixture_dirs = {
```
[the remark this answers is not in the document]
CostinTanasoiu: I can rewrite this test to go through all the files, parse each of them with and without the hint, and verify that we get the same results. This eliminates the need to know what the expected nature and format should be.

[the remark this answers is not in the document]
CostinTanasoiu: Rewritten!

spec/format_parser_spec.rb
CostinTanasoiu: There are many other places in this spec file that list specific natures and formats.

rcpalacio: @linkyndy I don't see an issue in FormatParser exposing a list of supported formats -- or natures. The parsers themselves can be isolated as internals, but the full support list can be interesting to consumers.

rcpalacio: @CostinTanasoiu the other ones I see as less of an issue because they are fine testing a subset. This one is tricky because it really wouldn't do it's job if someone creates a new parser in the future with a lower priority than mp3-- it only covers updates.

CostinTanasoiu: @rcpalacio I've rewritten the test. I exposed two new variables from FormatParser –`registered_natures` and `registered_formats` – and I'm using these to retrieve the parsers in my test. Let me know if this isn't ideal or if there's a better approach.

rcpalacio (approved):
👍

lib/parsers/mp3_parser.rb
linkyndy: Thanks for the details! And when they have the same prio, I would assume there's no definite ordering, right?

I'm just thinking whether anything that was after MP3s was kinda depending on this parser and now their behaviour might change 🤔 

If it's a non-issue, feel free to ignore 😊

lib/format_parser.rb
```diff
@@ -68,9 +68,22 @@ def self.register_parser(callable_parser, formats:, natures:, priority: LEAST_PR
       end
       @parser_priorities ||= {}
       @parser_priorities[callable_parser] = priority
+
+      @registered_natures ||= []
```
linkyndy, line 72: The `||=` doesn't really do what you'd expect here (not saying it doesn't work or that it isn't used above 😅 ). As a good practice, I'd see this below, when defining the reader, and then using the reader throughout the logic.

```
def self.registered_natures
    @registered_natures ||= []
end

# ...

registered_natures |= parser_provided_natures
```

(same for the other similar ones)

It also makes sense since you're anyway defining the reader for later access.

spec/format_parser_spec.rb
linkyndy: I think it looks good 👍

lib/parsers/mp3_parser.rb
CostinTanasoiu: It doesn't seem to be a big issue. When they have the same priority they end up in the order that they got registered. That will be the same every time unless we change something. In the future we might want to specify priorities to all of them. But for now after moving MP3 at the bottom, the tests seem to confirm that all parsers continue to function as expected.

lib/format_parser.rb
CostinTanasoiu: Refactored it.

lib/parsers/mp3_parser.rb
linkyndy: Awesome, thanks for confirming!

PR #241, opened 2023-09-14
Improved stability for mp4 parser when dealing with corrupted FTYP boxes

rcpalacio:
**About the issue**

Every MP4 file should start with a `ftyp` box. We rely on that to build a check for magic bytes and return early from the parser.

The problem arises when a file matches the header of a `ftyp` box but don't actually describe the box correctly -- for instance, bringing a wrong box size. In that case, the box can't be parsed and the file should be discarded.
What happens instead is that, due to an issue with the safe navigation, the parser would still try to read from the missing box, resulting in the error `NoMethodError: undefined method `[]' for nil:NilClass`.


spec/parsers/mp4_parser_spec.rb
```diff
@@ -31,6 +31,18 @@
...
     end
 
+    context "when when magic bytes are present but the 'ftyp' box can't be parsed" do
+      let(:result) do
+        input = [0xFF].pack('N') + 'ftyp' + 'avc1 ' + [0x1].pack('N')
+        io = StringIO.new(input)
+        subject.call(io)
+      end
+
+      it 'should be nil' do
```
linkyndy, line 41: I would keep the `subject` call in the test itself, and only prepare the "ingredients" for a test in a `let`, otherwise it feels like a scope creep between the two (`subject` should technically be the subject of a test, not of a _helper_).

linkyndy (approved):
Thanks for spotting and fixing this! 🔧

PR #243, opened 2024-06-28
Improve WAV parser

linkyndy:
Improve WAV parser by performing a best-effort when extracting metadata from files that do not strictly follow the format spec.

This deals with a particular case when non-PCM files might not have a `fact` chunk defined, in which case we currently discard the entire file. With this change, we will still recognise the file as a WAV file and attempt to build all relevant metadata from the available information in the header.


lib/parsers/wav_parser.rb
```diff
@@ -20,43 +20,39 @@ def call(io)
...
         # Skip this chunk until a known chunk is encountered
         safe_skip(io, chunk_size)
       end
+    rescue FormatParser::IOUtils::InvalidRead
+      # We've reached EOF, so it's time to make the most out of the metadata we
+      # managed to parse
+      break
     end
+
+    file_info(fmt_data, data_size, total_sample_frames)
```
RubenWolff-wt, line 47: Previously we returned from inside the loop depending on some conditions. Is there any issue with potentially having multiple chunks of the same type and then only returning for the final one of these?

lib/parsers/wav_parser.rb
```diff
@@ -20,43 +20,39 @@ def call(io)
...
     fmt_data = {}
-    total_sample_frames = 0
+    data_size = 0
+    total_sample_frames = nil
     loop do
       chunk_type, chunk_size = safe_read(io, 8).unpack('a4l')
       case chunk_type
       when 'fmt ' # watch out: the chunk ID of the format chunk ends with a space
         fmt_data = unpack_fmt_chunk(io, chunk_size)
-        return process_non_pcm(fmt_data, total_sample_frames) if fmt_data[:audio_format] != 1 and fact_processed
```
RubenWolff-wt, line 37: We no longer consider audio_format at all. Is that pretty much the fix we implemented?

lib/parsers/wav_parser.rb
```diff
@@ -20,43 +20,39 @@ def call(io)
...
-        return process_non_pcm(fmt_data, total_sample_frames) if fmt_processed and fmt_data[:audio_format] != 1
-        fact_processed = true
       else
         # Skip this chunk until a known chunk is encountered
         safe_skip(io, chunk_size)
       end
+    rescue FormatParser::IOUtils::InvalidRead
+      # We've reached EOF, so it's time to make the most out of the metadata we
+      # managed to parse
+      break
```
RubenWolff-wt, line 44: If we only break after the end of file is there potentially an issue when the file is very large? Should we potentially break when we have all the 3 expected chunks or at least any valid combination of them?

lib/parsers/wav_parser.rb
linkyndy: Indeed. The `fact` chunk is _kinda_ mandatory for non-PCM files, but the spec doesn't say it cannot be included in PCM files. Moreover, many non-PCM files don't have the `fact` chunk (the main issue we started this with). So yes, we'll be ignoring the `audio_format` and instead try to make the most of whatever chunks are provided.

lib/parsers/wav_parser.rb
linkyndy: As far as I got it, chunks can be unordered (more or less), but they cannot repeat themselves (at least, not the 3 ones referred to in this PR). That would lead to a pretty invalid file. And if that's the case, I don't think it'll matter much what we return.

lib/parsers/wav_parser.rb
RubenWolff-wt: A metric within peek seems fair enough (we probably already have something) 👍

lib/parsers/wav_parser.rb
RubenWolff-wt: Should we at least check that `fmt_data` exists as otherwise we access it inside file_info when it's potentially nil (which was previously impossible) and avoid random nil access errors?

RubenWolff-wt (approved):
lgtm

lib/parsers/wav_parser.rb
```diff
@@ -20,43 +20,39 @@ def call(io)
...
     end
+
+    file_info(fmt_data, data_size, total_sample_frames)
   end
 
   def unpack_fmt_chunk(io, chunk_size)
```
CostinTanasoiu, line 50: Does it still make sense to still extract the `audio_format` if you're no longer using it anywhere?
` audio_format: fmt_info[0]`

CostinTanasoiu (approved):

lib/parsers/wav_parser.rb
linkyndy: Good one. Refactored the area to get rid of that, and also of `block_align`, which was not used as well.

Kevin-McGonigle (approved):

PR #246, opened 2024-07-10
More improvements on WAV parser

linkyndy:
The changes recently introduced (v2.9.0) are too eager and forgiving, meaning that the reading budget almost always gets exceeded for large files. In this iteration we focus on:
* making sure we detect the file as WAV, regardless of how much metadata can be read
* ignoring the `fact` chunk and solely relying on `fmt ` and `data` to extract/compute the metadata
* reading as few bytes as possible rather than doing a thorough job extracing/computing **all** metadata according and not according to the format spec


rcpalacio (approved):
👍

RubenWolff-wt (approved):
lgtm

